PR #4, opened 2021-08-31
Projects download GMS with plots

SnowNooDLe:
# Projects download GMS with plots

### Removed unnecessary functions in seistech/utils.py
It does not mean I removed anything we used to have, I removed any functions I brought from the frontend to calculate to get x-axis and y-axis data. But there was a better way with numpy/pandas.

### Spectra plot in log scale
It was because we used a list of `string` as x-axis data instead of a list of `int`, thanks to Claudio.
gms_spectra_plot

### Mw Rrup plots with error bounds
gms_mw_rrup_plot

### Under project_api/utils.py
```python
split_gms_id = gms_id.split("_")
 disagg_dir = f"disagg_{split_gms_id[0]}_{split_gms_id[1]}_{int(1 / float(split_gms_id[2].replace('p', '.')))}"
```
Because when we generate disagg data for GMS, it creates a directory under `gms_path/disagg_data/{combination of im, rp}
image

For instance, the latest temporary projects with GMS (disagg data) path looks something like this
```
/{our_path}/projects/v21p8p1/test/results/Arthurs_Pass_200_0p045_0p275/gms_pSA_0p3_0p0004_PGA_pSA/disagg_data/disagg_pSA_0p3_2500
```
The code above, `disagg_dir` is to get this part `disagg_data/disagg_pSA_0p3_2500` after the disagg_data/.

Projects download data now contains all the plots in png format.

image

## Log scale issue is fixed now

Currently, it's not on the log scale.

### GMS Spectra with log scale
gms_spectra_plot_log

### GMS Spectra with log scale & limit
gms_spectra_plot_log_limit

### GMS Spectra without log scale
gms_spectra_plot


jasonmotha:
Yes, we do want it log scale. For your reference what you've done on the "without log scale" plots, you've just plotted each data point rather than have a continuous scale on the axis (the difference between each point is not even)

calculation/seistech_calc/seistech_calc/plots/plt_plotting.py
```diff
@@ -549,3 +592,403 @@ def plot_uhs_branches(
...
+    ----------
+    gms_result_data: Dict,
+    save_file: Path, optional
+    """
+    plots_data = utils.calculate_gms_im_distribution(gms_result_data)
+
+    plt.figure(figsize=(16, 9))
+
+    def label_maker(im: str):
+        return (
```
claudio525, line 613: Don't think its worth defining a function for this if it only used in one location?

calculation/seistech_calc/seistech_calc/plots/plt_plotting.py
```diff
@@ -549,3 +592,403 @@ def plot_uhs_branches(
...
+        plt.show()
+
+
+def plot_gms_available_gm(
+    metadata: Dict,
+    bounds: Dict,
+    num_in_bounds: int,
+    save_file: Path = None,
+):
+    """GMS Available ground motions plot
```
claudio525, line 941: Can we extend the docstrings for these  functions a bit more, i.e. be more specific in terms of what plot is produced, e.g.
`Plot of GMs in the GM dataset with respect to distance (Rrup) and Magnitude (Mw). Also shows the causal parameters used during ground motion selection`

claudio525: If you are not sure about what to put, post a picture of the plot along with the function name on the slack channel and I can help you with an appropriate description.

calculation/seistech_calc/seistech_calc/plots/plt_plotting.py
```diff
@@ -549,3 +592,403 @@ def plot_uhs_branches(
         plt.close()
     else:
         plt.show()
+
+
+def plot_gms_im_distribution(
+    gms_result_data: Dict,
```
claudio525, line 598: Also I suggest using the GMS result objects directly instead of dictionary derived from it. (I know the other plotting functions also do this, however since the plots are now part of `seistech_calc` and are only going to be used with `seistech_calc` result objects, we should switch to just working on those. )

Sorry, probably should have realised that sooner..

calculation/seistech_calc/seistech_calc/utils.py
```diff
@@ -418,3 +420,244 @@ def to_mu_sigma(df: pd.DataFrame, im: IM):
     return df.loc[:, [str(im), f"{im}_sigma"]].rename(
         columns={str(im): "mu", f"{im}_sigma": "sigma"}
     )
+
+
+def calculate_gms_spectra(gms_data: Dict, num_gms: int):
```
claudio525, line 425: I'm not 100% sure what this function does, however it seems quite complex. Might be worth for us to go through this together and see if we can simplify it.

claudio525: I assume its a pretty direct port from the javascript frontend code?

claudio525 (commented):
Nice work, just couple of things to tidy up and maybe some simplification (lets discuss)

jasonmotha (changes requested):

calculation/seistech_calc/seistech_calc/constants.py
```diff
@@ -74,3 +74,28 @@ class GMSourceType(Enum):
     historical = "historical"
     simulations = "simulations"
     mixed = "mixed"
+
+
+class GMSIMDistributionsLabel(Enum):
```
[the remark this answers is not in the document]
joelridden: My opinion would be to do the dictionary mapping. If this was used in many different files then adding it to the IM class would be more useful

joelridden (dismissed):

jasonmotha (approved):

calculation/seistech_calc/seistech_calc/plots/plt_plotting.py
```diff
@@ -549,3 +618,479 @@ def plot_uhs_branches(
...
+):
+    """CDF plots for the selected GMs and
+    all ruptures (disaggregation) for Magnitude
+    and Distance (Rrup)
+
+    Also shows causal bounding parameters if specified
+
+    Parameters
+    ----------
+    contribution: List
```
claudio525, line 978: Both the `contribution` and `distribution` parameter are unclear as to what they are?
Please specify where they come from, or preferably use the result objects, i.e. GMSResult and DisaggResult

claudio525: Also whats the difference between `gms_metadata` and `metadata`

calculation/seistech_calc/seistech_calc/plots/plt_plotting.py
```diff
@@ -549,3 +618,479 @@ def plot_uhs_branches(
...
+
+    if save_file is not None:
+        plt.savefig(save_file)
+        plt.close()
+    else:
+        plt.show()
+
+
+def plot_gms_available_gm(
+    metadata: Dict,
```
claudio525, line 1035: Use GMSResult if possible pls

calculation/seistech_calc/seistech_calc/utils.py
```diff
@@ -418,3 +422,96 @@ def to_mu_sigma(df: pd.DataFrame, im: IM):
     return df.loc[:, [str(im), f"{im}_sigma"]].rename(
         columns={str(im): "mu", f"{im}_sigma": "sigma"}
     )
+
+
+def calculate_gms_spectra(
```
claudio525, line 427: This function is only used by the `plot_gms_spectra` function, therefore it should move into the plt_plotting module and become protected (i.e. start with _)
Also something like `prepare_gms_spectra` might be more accurate (since its not actually calculating anything)

calculation/seistech_calc/seistech_calc/plots/plt_plotting.py
```diff
@@ -549,3 +618,479 @@ def plot_uhs_branches(
...
+        sorted_realisations = realisations.loc[:, str(IMi)].sort_values()
+        plt.step(
+            sorted_realisations.values,
+            np.linspace(0, 1, len(sorted_realisations)),
+            where="post",
+            color="blue",
+            label="Realisations",
+        )
+
+        sorted_selected_gms = selected_gms.loc[:, str(IMi)].sort_values()
```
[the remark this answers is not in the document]
SnowNooDLe: I can certainly replace to something like this
```python
plt.step(
    selected_gms.loc[:, str(IMi)].sort_values().values,
    np.linspace(0, 1, len(selected_gms.loc[:, str(IMi)])),
    where="post",
    color="black",
    label="Selected Ground Motions",
)
```
I just didn't like the way it gets longer within step's arguments haha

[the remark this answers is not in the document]
claudio525: Thats what I would do, but its kinda a preference thing. I personally don't like having variables that are only used (unless the expression that defines them is massive). Up to you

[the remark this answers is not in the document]
SnowNooDLe: Actually, kinda makes sense, unless it needs to be used somewhere else, doesn't really have to be a tidy looking variable :thinking:

calculation/seistech_calc/seistech_calc/plots/plt_plotting.py
SnowNooDLe: Added some extra explanation for `contribution` and `distribution`.
Also instead of using `gms_metadata`, accepting gmsResult object with `metadata_key` instead of `metadata`

calculation/seistech_calc/seistech_calc/plots/plt_plotting.py
```diff
@@ -731,11 +729,17 @@ def plot_gms_mw_rrup(
 
     Parameters
     ----------
-    metadata: Dict
+    gms_result: gms.GMSResult
     disagg_mean_values: pd.DataFrame
```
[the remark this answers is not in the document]
SnowNooDLe: Nice catch, is supposed to be series :joy:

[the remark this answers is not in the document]
SnowNooDLe: >Also, I'm not a fan of having this as a docstring if it's just repeating the type-hinting / default arguments above as it's just two things we need to keep in-sync.

This is a fair call as some docstring like this, with no extra explanation, is basically duplicate information from the function's param. E.g., def plot_gms_mw_rrup(gms_result: gms.GMSResult,....), we know that `gms_result` is an GMSResult object.

However, because we've been keeping parameters inside the docstring regardless, will keep it for now.
Perhaps, we can also have a chat about this like we did for the import guidelines.

jasonmotha (dismissed):

joelridden (dismissed):

calculation/seistech_calc/seistech_calc/plots/plt_plotting.py
```diff
@@ -549,3 +618,610 @@ def plot_uhs_branches(
...
+    gms_result: gms.GMSResult
+    metadata_key: str
+    cs_param_bounds: gms.CausalParamBounds, optional
+    save_file: Path, optional
+    """
+    selected_gms_metadata = gms_result.selected_gms_metdata_df.to_dict(orient="list")
+    bounds = _get_causal_params_bounds(cs_param_bounds)
+    bounds_y_range = [0, 1]
+    # To achieve Empirical distribution function with matplotlib's step plotting
+    copied_metadata = selected_gms_metadata[metadata_key][:]
```
claudio525, line 1002: I suggest changing these lines to the following (only changed the first one, and renamed variable)
```
    metadata_values = list(gms_result.selected_gms_metdata_df.loc[:, metadata_key].values)
    metadata_values.append(min(copied_metadata))
    metadata_values.sort()

```
that way you can remove the line at 998

calculation/seistech_calc/seistech_calc/plots/plt_plotting.py
```diff
@@ -549,3 +618,610 @@ def plot_uhs_branches(
...
+
+    if save_file is not None:
+        plt.savefig(save_file)
+        plt.close()
+    else:
+        plt.show()
+
+
+def plot_gms_disagg_distribution(
+    contribution: List,
```
claudio525, line 973: Instead of passing in the x and y values as two variables I suggest that this function takes a pd.Series, where the index are the x-values (i.e. either the mag or rrup values) and the values are the corresponding contribution values.

tools/seistech_utils/seistech_utils/src/api/shared.py
```diff
@@ -417,6 +420,132 @@ def create_uhs_download_zip(
...
+
+        # IM distribution plots
+        sc.plots.plt_gms_im_distribution(gms_result, save_file=Path(out_dir))
+
+        # Disagg Distribution plots (Mw Distribution or Rrup distribution)
+        contribution_df_data = sr.get_default_causal_params(cs_param_bounds)[
+            "contribution_df"
+        ]
+        if contribution_df_data is not None:
+            sorted_rrup, rrup_cdf = utils.calc_cdf(
```
claudio525, line 464: Instead of doing the sorting here and the plotting function assuming that the passed in data is sorted, I suggest that the sorting is moved into the plotting function. Combined with the other comment above of taking a series instead of 2 separate lists the call from here then just becomes the following (for magnitude)
```
            sc.plots.plt_gms_disagg_distribution(
               cs_param_bounds.contr_df.loc[:, ["magnitude", "contribution"]].set_index("magnitude", drop=True),
                gms_result,
                "mag",
                cs_param_bounds=cs_param_bounds,
                save_file=Path(out_dir)
                / f"{prefix}gms_mag_disagg_distribution_plot.png",
            )
```
And you can remove line 460 & 464-470

claudio525 (commented):
Couple of more comments. Let me know if anything isn't clear, happy to clarify further!

calculation/seistech_calc/seistech_calc/plots/plt_plotting.py
```diff
@@ -1094,3 +1114,114 @@ def plot_gms_available_gm(
...
+    selected_gms_df = selected_gms_df.T.sort_index().T
+
+    return (
+        gcim_df,
+        realisations_df,
+        selected_gms_df,
+    )
+
+
+def _get_causal_params_bounds(cs_param_bounds: gms.CausalParamBounds):
```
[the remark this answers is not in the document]
SnowNooDLe: Under seistech_utils/api/shared_response.py, we already have a function that returns all the cs_param_bounds' low & high.
Which is returning raw data, no modification at all to the frontend, this _get_causal_params_bounds is a slightly modified version as the variable name is not consistent,(E.g., some places we use Mw, some places mag) hence I created this function within this `plot_plotting.py` to get min/max value a bit easy.

Perhaps I can move the following function to the CasualParamsBounds class as a method but not sure about moving `_get_causal_params_bounds()` as this is really specialized function just for plotting for us to get value a bit easy.

Happy to discuss if needed!

```python
def get_default_causal_params(cs_param_bounds: sc.gms.CausalParamBounds):
    """Creates the response for both the core and Project API"""
    return {
        "mw_low": cs_param_bounds.mw_low,
        "mw_high": cs_param_bounds.mw_high,
        "rrup_low": cs_param_bounds.rrup_low,
        "rrup_high": cs_param_bounds.rrup_high,
        "vs30_low": cs_param_bounds.vs30_low,
        "vs30_high": cs_param_bounds.vs30_high,
        "contribution_df": cs_param_bounds.contr_df.to_dict("list")
        if cs_param_bounds.contr_df is not None
        else None,
    }
```

tools/seistech_utils/seistech_utils/src/api/shared.py
```diff
@@ -417,6 +420,132 @@ def create_uhs_download_zip(
     return zip_ffp
 
 
+def write_gms_download_data(
+    gms_result: sc.gms.GMSResult,
+    app: flask.app,
```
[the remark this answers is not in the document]
SnowNooDLe: Glad python does returning multiple values.

calculation/seistech_calc/seistech_calc/plots/plt_plotting.py
SnowNooDLe: Good to practice more of DataFrame then converting them to dictionary :+1:

tools/seistech_utils/seistech_utils/src/api/shared.py
SnowNooDLe: Did manage to sort within the plotting function. However, had to bring `calc_cdf` function to `plt_plotting.py` due to circular imports.

jasonmotha (commented):
Can't see any issues with code. Can you go into a bit more about how the circular import was created?

Does plotting depend on utils or does utils depend on plotting or both?

calculation/seistech_calc/seistech_calc/plots/plt_plotting.py
```diff
@@ -1225,3 +1237,21 @@ def _get_causal_params_bounds(cs_param_bounds: gms.CausalParamBounds):
             "vs30": cs_param_bounds.site_info.db_vs30,
         },
     }
+
+
+def _calc_cdf(weights: List, x_values: List):
```
claudio525, line 1242: Don't think its worth making a function for this if its only used once.
Also you can sort a dataframe by its columns directly, i.e. in line 1005 becomes
`contribution_df = contribution_df.sort_values()` and then lines 1019 and 1020 become
`contribution_df.index.values` & `contribution_df.values`
Also I suggest renaming `contribution_df` to `contribution_series` since its actually a pd.Series not pd.DataFrame 
Also passing numpy arrays to matplotlib is perfectly fine, so no need to cast to list as you do in line 1019 & 1020

SnowNooDLe:
> Can't see any issues with code. Can you go into a bit more about how the circular import was created?
> 
> Does plotting depend on utils or does utils depend on plotting or both?

The function `calc_cdf` was used to be inside the `seistech_utils` as I was planning to use within `seistech_utils/shared.py`.
(`seistech_utils` imports `seistech_calc`)
Then import `calc_cdf` to somewhere else, E.g. as suggested, to the plotting function which is sitting inside the `seistech_calc`, then I had to import `seistech_utils` from `seistech_calc` which caused circular import.

jasonmotha:
> > Can't see any issues with code. Can you go into a bit more about how the circular import was created?
> > Does plotting depend on utils or does utils depend on plotting or both?
> 
> The function `calc_cdf` was used to be inside the `seistech_utils` as I was planning to use within `seistech_utils/shared.py`.
> (`seistech_utils` imports `seistech_calc`)
> Then import `calc_cdf` to somewhere else, E.g. as suggested, to the plotting function which is sitting inside the `seistech_calc`, then I had to import `seistech_utils` from `seistech_calc` which caused circular import.

Makes sense.

@claudio525 as part of our stabilization phase, it might be a good idea to get a pictorial map of our imports and have a look if everything is laid out roughly how we expect. I think there might be a couple of functions in the wrong package etc.

calculation/seistech_calc/seistech_calc/plots/plt_plotting.py
SnowNooDLe: Apparently, it is actually `pd.DataFrame`

As suggested, used the following code which is still a DatFrame.
```python
cs_param_bounds.contr_df.loc[
    :, ["contribution", "magnitude"]
].set_index("magnitude", drop=True)
```                

image

joelridden (approved):

jasonmotha (approved):

claudio525 (approved):

PR #5, opened 2021-09-01
Import seistech frontend repo

SnowNooDLe:
# Import the seistech frontend repo

## This is more about import&export, no need code review as these are from the frontend repo.

Upcoming PR is about making the frontend and intermediate API work. Such as
- Rename middleware to intermediate API
- Update Dockerfile to work in the updated folder structure
    - Install packages from the seistech repo, not from seistech frontend


jasonmotha (approved):
Hmmm, you might actually be earning your lines of code :rofl: 

Thanks for the no need to review notice.

claudio525 (approved):

PR #10, opened 2021-09-05
Updated ERF index

claudio525:
Converted ruptured dataframe to use an index instead of rupture id to reduce memory usage for ensembles with a large number of branches.

- All rupture dataframes stored in Branches, IMEnsembles and Ensemble now have an int index  as df index instead of rupture id
- Conversion between rupture_id -> rupture_id_ix (and the other way) is done via a lookup in the ensemble 
- For ease of use Branches, IMEnsembles and Ensemble still expose the rupture id dataframes via a property, but obviously this now involves an on the fly lookup


calculation/seistech_calc/seistech_calc/disagg/DisaggData.py
```diff
@@ -58,30 +59,47 @@ def __init__(
...
+    @property
+    def ds_disagg_id(self):
+        raise NotImplementedError()
+
     @property
     def total_contributions(self) -> pd.Series:
-        return self.fault_disagg.contribution.append(
-            pd.Series({"distributed_seismicity": self.ds_disagg.contribution.sum()})
+        return self.fault_disagg_id.contribution.append(
+            pd.Series({"distributed_seismicity": self.ds_disagg_id.contribution.sum()})
```
SnowNooDLe, line 91: Please correct me if I understood something wrong, I've seen @property decorators a few times but wasn't sure what exactly does, especially when it does something more than like what these two do, obviously these two seems just getters. [ref](https://www.programiz.com/python-programming/property) and by reading this, it seems basically it's a getter/setter pythonic way.
```python
property
    def fault_disagg_id_ix(self):
        return self._fault_disagg

    @property
    def ds_disagg_id_ix(self):
        return self._ds_disagg
```

Anyway, if so, wouldn't this be an error?
`self.fault_disagg_id.contribution.append` because `self.fault_disagg_id` is basically the property decorator right above, `def fault_disagg_id(self)` which is currently raising not implemented error?

Was it meant to be something like this?
```python
return self.fault_disagg_id_ix.contribution.append ...... "distributed_seismicity": self.ds_disagg_id_ix.contribution.sum()})
```

calculation/seistech_calc/seistech_calc/disagg/DisaggData.py
```diff
@@ -222,6 +258,24 @@ def __init__(
...
+        )
+
+    @property
+    def ds_disagg_id(self):
+        return self._ds_disagg.set_index(
+            rupture.rupture_id_ix_to_rupture_id(
+                self.im_ensemble.ensemble, self._ds_disagg.index.values
+            ),
+            inplace=False,
+        )
```
joelridden, line 277: Do these need to be here? Can't we just use DisaggData's properties as they seem the same to me? No need to override them.

calculation/seistech_calc/seistech_calc/disagg/disagg.py
```diff
@@ -360,15 +378,13 @@ def run_disagg_gridding(
     )[:, 0]
     ds_ruptures["rupture_id"] = ds_ruptures.index.values
 
-    # Get the distances
+    # Add distance data
```
joelridden, line 381: Still seems to be a get here rather than an addition?

calculation/seistech_calc/seistech_calc/gm_data/Ensemble.py
```diff
@@ -1,12 +1,13 @@
 import os
 import hashlib
 from glob import glob
-from typing import Dict, TYPE_CHECKING
+from typing import Dict, TYPE_CHECKING, Sequence
```
joelridden, line 4: I don't see where Sequence was used here so can probably be safely removed

calculation/seistech_calc/seistech_calc/rupture/rupture.py
```diff
@@ -74,3 +80,29 @@ def rupture_name_to_id(rupture_names: np.ndarray, erf_ffp: str):
...
+):
+    """Convertes the rupture names to rupture_id_ix values"""
+    return rupture_id_to_ix(
+        ensemble, rupture_name_to_id(rupture_names, erf_ffp)
+    )
+
+
+def rupture_id_ix_to_rupture_id(ensemble: "gm_data.Ensemble", rupture_id_ind: np.ndarray):
+    """Converts rupture id ix to rupture ids"""
+    return ensemble.get_rupture_ids(rupture_id_ind)
```
joelridden, line 108: No new line end of file

calculation/seistech_calc/seistech_calc/disagg/DisaggData.py
jasonmotha: As is it looks like it will raise a non implemented error. I'm happy to be educated otherwise.

SnowNooDLe: Great, I knew that was the case!
Thanks for the clarification :+1:

calculation/seistech_calc/seistech_calc/disagg/DisaggData.py
SnowNooDLe: Agreed, it seems they are identical(BranchDisaggData and EnsembleDisaggData) as well, I reckon it can be moved to the parent's class.

calculation/seistech_calc/seistech_calc/disagg/DisaggData.py
claudio525: It doens't actually result in a `NotImplementedError` as a DisaggData class should never be instantiated directly. Instead only BranchDisaggData and EnsembleDisaggData are instantiated, which both implement `fault_disagg_id` and `ds_disagg_id`. 

I have added a comment to these to indicate that these have to implement by sub-classes

jasonmotha: Oh yep. That makes sense. Would it be possible to mark these as abstract methods / class? I think python has a module for that?

calculation/seistech_calc/seistech_calc/disagg/DisaggData.py
claudio525: THere is a difference between them in the way that they access the ensemble, which is needed for the lookup. However, I should have at least removed the duplication...
But moving it to DisaggData is even better (but it didn't use to have access to the ensemble, and I didn't think it was a good idea to add it, but have changed my mind on that..). Anyways properties will move to DisaggData as suggested.

calculation/seistech_calc/seistech_calc/disagg/DisaggData.py
claudio525: Hmh pretty sure I looked at that once and decided for some reason against it. Not sure, not keen on doing this now (since we don't do it anywhere else). But I do agree that it should be more obvious, I'll rename it to `BaseDisaggData` which I think is what we do with other pure BaseClasses.

joelridden (approved):

jasonmotha (approved):

SnowNooDLe (approved):

PR #13, opened 2021-09-10
Cs 21p6 changes

joelridden:
- A few changes to im csv to imdb to allow for specifying a component.

- Fix in utils with importing os

- Updates to some configs

### DIFF
--- a/tools/seistech_utils/seistech_utils/src/utils.py
+++ b/tools/seistech_utils/seistech_utils/src/utils.py
@@ -1,3 +1,5 @@
+import os
+
 from pathlib import Path
 from typing import Tuple, Union, Sequence, List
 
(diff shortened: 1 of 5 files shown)


tools/seistech_utils/seistech_utils/src/utils.py
```diff
@@ -1,3 +1,5 @@
+import os
+
```
SnowNooDLe, line 2: You do not need an extra line between `import os` and `from pathlib....` they are all Python built-in stuff. :+1:

SnowNooDLe (dismissed):
Nice and simple, unlike my recent PR... :joy:

SnowNooDLe (approved):

jasonmotha (approved):

claudio525 (approved):

PR #15, opened 2021-09-26
Lots of fixes 

claudio525:
+ Added new aggregated disagg data file to project zip
+ Removed src folder in sha_calc


calculation/sha_calc/sha_calc/gcim/gcim_emp.py
```diff
@@ -6,8 +6,8 @@
 import numpy as np
 from scipy import stats
 
-import sha_calc.src.disagg as disagg
-import sha_calc.src.ground_motion as gm
+from sha_calc import disagg
+import sha_calc.ground_motion as gm
```
SnowNooDLe, line 10: Let's swap the lines
`import` to be at the top

calculation/seistech_calc/seistech_calc/disagg/DisaggResult.py
```diff
@@ -288,13 +269,25 @@ def get_save_dir(im: IM, exceedance: float = None, im_value: float = None):
...
     def load(cls, data_dir: Path):
-        metadata, site_info, fault_disagg, ds_disagg, mean_values = cls._load_data(
-            data_dir
+        with open(data_dir / cls.METADATA_FN, "r") as f:
+            metadata = json.load(f)
+
+        mean_values = (
+            pd.read_csv(data_dir / cls.MEAN_VALUES_FN, index_col=0).squeeze()
+            if (data_dir / cls.MEAN_VALUES_FN).exists()
+            else None
```
joelridden, line 278: Not sure if this really needs to change? I see that it's only used here currently, but in future if we did want to implement BranchDisaggResult load wouldn't we still want that base function.

tools/seistech_utils/seistech_utils/src/utils.py
```diff
@@ -2,12 +2,20 @@
...
+        # Available for python version >= 3.8
+        from importlib.metadata import version
+    except ImportError:
+        try:
+            # Third party package
+            # https://pypi.org/project/importlib-metadata/
+            from importlib_metadata import version
+        except ImportError:
+            print(f"Please install the importlib-metadata package "
+                  f"or switch to a python version >= 3.8")
```
jasonmotha, line 16: This feels messy. But looks like the best way to do it.

jasonmotha (approved):
Pending others comments too.

tools/seistech_utils/seistech_utils/src/utils.py
claudio525: It is, I just added it for running on Maui, but since I gave up on that it probably makes sense to just remove it.

calculation/seistech_calc/seistech_calc/disagg/DisaggResult.py
claudio525: Yes, and if we do then it should be changed back. But we haven't needed to implement a BranchDisaggResult load function at this point, so it probably won't ever be needed. So then the base function just becomes extra spaghetti code imo

SnowNooDLe (approved):

calculation/seistech_calc/seistech_calc/disagg/DisaggResult.py
joelridden: Ok fair

joelridden (approved):

jasonmotha (approved):

PR #18, opened 2021-10-12
Directivity Part 2

joelridden:
### **Directivity Part 2 - Hypocentre Averaging**

Note: Some of this would've already been reviewed in Part 1, doesn't seem to be showing the actual lines changed for Part 2

- Averaging bea20 model outputs by using hypocentre adjustment across strike and depth of the fault
- Adding validation plot function
- Fixes for multi-plane faults
- Updated benchmark test data

Alpine Example output plot with 20 hypocentres for averaging:
validation_plot_AlpineK2T

In next Part/PR:
- Use of virtual srfs with GMHazard integration


jasonmotha:
I think you should be able to merge main into your branch to tidy up the differences.

jasonmotha:
Actually, you also haven't pushed anything to this branch in the last 19 days :thinking: Maybe you just need to push?

calculation/sha_calc/sha_calc/directivity/bea20/utils.py
```diff
@@ -67,11 +67,8 @@ def set_hypocentres(n_hypo: int, planes: List, depth_method: List):
         current_length = 0
         planes_copy = [plane.copy() for plane in planes]
         for index, plane in enumerate(planes_copy):
-            if (
-                current_length < distance
-                and (current_length + plane["length"]) > distance
-            ):
-                plane["shyp"] = distance - (current_length + plane["length"] / 2)
+            if current_length < distance < (current_length + plane["length"]):
+                plane["shyp"] = distance - (total_length / 2)
```
SnowNooDLe, line 71: Just checking, line 70 is just a simplified version
but line 71 changed from 
```python
plane["shyp"] = distance - (current_length + plane["length"] / 2)
```
to 
```python
plane["shyp"] = distance - (total_length / 2)
```
Is this meant to be happened?

joelridden: Yes this was meant to happen, the shyp value was meant to be from the centre of the plane it was in, but turns out it was actually from the centre of the entire fault and so this adjustment was needed so that GC2 gets the correct location for the hypocentre.

calculation/sha_calc/sha_calc/directivity/bea20/directivity.py
```diff
@@ -31,45 +32,363 @@ def compute_directivity_effects(
...
+        fd_average,
+        fdi_average,
+        phi_red_average,
+        phi_redi_average,
+        predictor_functions_average,
+        other_average,
+    )
+
+
+def directivity_plots(
```
claudio525, line 145: I would suggest making the validation plotting functionality scripts instead of functions, 
- 1 script for generating the non-hypocentre validation plots (i.e. the one with 6 sub-figures)
- 1 script for generating the hypocentre-averaring plots (both average & individiuals, but only fdi (unless there is a good reason to plot the others?))
- Put shared plotting functionality into a plots.py module

claudio525: And put said scripts along with the output plots into a `validation` folder

calculation/sha_calc/sha_calc/directivity/bea20/directivity.py
```diff
@@ -10,7 +10,7 @@
 from sha_calc.directivity.bea20 import bea20, utils
 
 
-def compute_directivity_single_hypo(
+def compute_directivity_srf_single(
     srf_file: str, srf_csv: Path, sites: np.ndarray, period: float = 3.0
 ):
     """Computes directivity effects at the given sites with the given srf data for a single hypocentre
```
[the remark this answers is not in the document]
joelridden: Was more single for a single hypocentre, that is at the end of the docstring. Maybe adding hypo at the end of the function could make it more understandable? Just don't want the function name to be too long.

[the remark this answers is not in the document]
SnowNooDLe: Fair call, it probably becomes a bit longer if we add `hypo` at the end.

claudio525 (approved):

PR #19, opened 2021-10-13
Create projects DBs with extra models from open-quake

SnowNooDLe:
# Create projects DBs with extra models from open-quake

## Aware of the potential file names changed, just opening PR to provide an idea of whats changed

- Creating new projects with the following data:
    - Extra models from open-quake are used 
    - Two stations were used, one from the Christchurch(CCCC) and another one from the Wellington(TEPS)
    - Six IMs were used, PGA, PGV, pSA 0.1, 1.0, 3.0 and 7.5
    - No Disaggregation and UHS for now

- Update on the usage of function, `is_any_strings_in_string`
As long as we keep our naming convention, this should work, the reason to modify this was due to having open-quake model and one that's specialized for NZ.
For instance, `K_20_SS` and `K_20_SS_NZ` `K_20` is from open-quake, `kuehn_2020` and by doing `find_matching_dbs`, it will return a list with two elements due to checking `K_20_SS` in DB list which would raise an assertion by
```python
assert (
  0 <= len(ds_dbs) <= 1
  and 0 <= len(flt_dbs) <= 1
  and len(ds_dbs) + len(flt_dbs) >= 1
), f"{ds_dbs}, {flt_dbs}, {tect_type}, {model}"
```
Hence instead of checking the model by itself, combined with tectonic type as with our naming convention, ds/flt file is created in the following format, 
- {MOEL}_{TECT_TYPE}_{ds or flt}.db

- Minor change, while loading NZ code data, if the IM is neither pSA nor PGA, instead of raising an error, set them as `None` as there is nothing to display if the IM is not one of them(pSA or PGA)


tools/project_gen/project_gen/project.py
```diff
@@ -318,7 +318,7 @@ def generate_dbs(
         "--z-file",
         str(z_ffp),
     ]
-    ds_timeout = (n_stations * (60 * 60)) / (min(n_procs - 1, n_stations))
+    ds_timeout = (n_stations * (3600 * 24)) / (min(n_procs - 1, n_stations))
```
jasonmotha, line 321: This is probably too high? What's the expected time for one station to calculate?

tools/project_gen/project_gen/scripts/temp.yaml
```diff
@@ -0,0 +1,14 @@
+id: "temp_hazard_curve_two_sites_new_weights"
```
jasonmotha, line 1: Unsure if this should be commited.

jasonmotha: I mean in the sense that we generally create these yaml files as we want create projects. I don't think we'd want to run this specific scenario in the future.

tools/project_gen/project_gen/package_mapping.yaml
```diff
@@ -41,6 +41,18 @@ pga_psa:
   uhs_return_periods: [ 25, 100, 150, 250, 500, 1000, 2500, 5000, 10000 ]
   im_components: ["RotD50", "Larger"]
 
+pga_psa_pgv_temp:
```
claudio525, line 44: If this is temp, then it probably shouldn't be added to git?

tools/project_gen/project_gen/project.py
```diff
@@ -111,7 +112,7 @@ def create_project(
                 / "db_creation"
                 / "empirical_db"
                 / "empirical_model_configs"
-                / "20p4.yaml"
+                / "21p10.yaml"
```
claudio525, line 115: Thoughts on making this a function argument instead of hardcoding it? @jasonmotha

jasonmotha: Happy with either tbh. Chances are we will only have one active version (especially in an automated sense) and the other reason to change it would be testing a new config.

tools/project_gen/project_gen/project.py
```diff
@@ -36,6 +36,7 @@ def create_project(
     erf_pert_dir: Path = None,
     flt_erf_version: str = "NHM",
     setup_only: bool = False,
+    empirical_model_config: str = "21p10.yaml",
```
jasonmotha, line 39: I'd make this a constant in one of the files and then import references to it.

tools/project_gen/project_gen/project.py
```diff
@@ -192,7 +197,9 @@ def write_project_config(project_dir: Path, project_specs: Dict):
         project_config = yaml.safe_load(f)
 
         if project_specs.get("project_parameters") is not None:
-            print("Project parameters are already specified  set, skipping package type mapping")
+            print(
+                "Project parameters are already specified  set, skipping package type mapping"
```
jasonmotha, line 201: double space between specified and set -- I don't think it makes sense either.

tools/project_gen/project_gen/project.py
```diff
@@ -24,6 +24,8 @@
     "CFM_v21p8p1": "CFM_v0_9_Akatore_mod_21p8p1",
 }
 
+EMPIRICAL_MODEL_CONFIG_VERSION = "21p10"
```
jasonmotha, line 27: Just being difficult, but from a usability perspective, it's nice to pass in the full file path rather than a version that then has to be in a specific folder.

PR #25, opened 2021-11-04
MP support for DS IMDB generation

claudio525:


tools/gmhazard_scripts/db_creation/empirical_db/common.py
```diff
@@ -241,71 +313,30 @@ def calculate_emp_site(
...
+                for index, rupture in matching_df.iterrows()
+            ],
+        )
 
+    im_result_dict = {key: {} for key in imdb_dict.keys()}
+    for cur_rupture_id, cur_fault_im_dict in results:
+        for cur_gmm, cur_im_dict in cur_fault_im_dict.items():
+            im_result_dict[cur_gmm][cur_rupture_id] = cur_im_dict
+
+    im_result_df_dict = {key: {} for key in imdb_dict.keys()}
```
jasonmotha, line 339: I think this line isn't meant to be there.

tools/gmhazard_scripts/db_creation/empirical_db/common.py
```diff
@@ -241,71 +313,30 @@ def calculate_emp_site(
...
-        for imdb_key in imdb_dict.keys():
-            im_result_dict[imdb_key][rupture_id] = fault_im_result_dict[imdb_key]
+                for index, rupture in matching_df.iterrows()
+            ],
+        )
 
+    im_result_dict = {key: {} for key in imdb_dict.keys()}
+    for cur_rupture_id, cur_fault_im_dict in results:
+        for cur_gmm, cur_im_dict in cur_fault_im_dict.items():
+            im_result_dict[cur_gmm][cur_rupture_id] = cur_im_dict
```
jasonmotha, line 337: I remember the below df.from_dict being slow too. So if that is out of the mp pool there will be some waiting while that gets done. 

Short of creating a fixed sized df beforehand, I can't think of anything.

jasonmotha (changes requested):
Looks pretty good. Do you have benchmarks for before and after?

tools/gmhazard_scripts/db_creation/empirical_db/common.py
```diff
@@ -241,71 +313,30 @@ def calculate_emp_site(
...
+        results = p.starmap(
+            __process_rupture,
+            [
+                (
+                    rupture,
                     site,
-                    GMM,
-                    str(im_type),
-                    psa_periods if im_type is IMType.pSA else None,
+                    rupture_df,
```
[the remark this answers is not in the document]
joelridden: If this value is calculated outside of the __process_rupture function you might as well just set that as the first value in the tuple of (rupture_id, fault_im_result_dict) and just let __process_rupture return fault_im_result_dict. As it's not really used in the function at all that I can see.

tools/gmhazard_scripts/db_creation/empirical_db/common.py
claudio525: Meh it might slow, but is only done a couple of times. There might be a faster way, but I don't think its worth it.
From memory, when I ran it with MP the slowest part by far is still the loop over ruptures (which now uses MP)

jasonmotha: Oh yeah. It's only run once per station. Which isn't much for your use case.

jasonmotha (dismissed):

tools/gmhazard_scripts/db_creation/empirical_db/common.py
```diff
@@ -320,19 +314,21 @@ def calculate_emp_site(
...
                     site,
-                    rupture_df,
                     im_types,
                     tect_type_model_dict,
                     psa_periods,
                     keep_sigma_components,
                     {key: {} for key in imdb_dict.keys()},
                 )
-                for index, rupture in matching_df.iterrows()
+                for index, rupture in matching_df.iloc[:10].iterrows()
```
joelridden, line 323: Is there a reason why only the first 10 ruptures are looked at? or was this just testing?

SnowNooDLe (approved):
All good with me when they are both okay as I was waiting for updates on their feedback :+1:

tools/gmhazard_scripts/db_creation/empirical_db/common.py
claudio525: Good catch, that was testing

joelridden (approved):

SnowNooDLe (approved):

PR #30, opened 2021-11-16
Permission check

SnowNooDLe:
# Permission check

- Bring back a single `proxy_to_api` function that only deals with forwarding requests to the right destination.
- Permission check is happening within the Intermediate API
    - Before proxy to Core/Project API
    - Every intermediate API calls except one call

## Core API
- Must be authenticated via Auth0
- Must have certain permission with Auth0 to use, such as
    - hazard - To be able to access to non-projects tab and Site Selection
    - hazard:hazard - To use Hazard Curve
    - hazard:disagg - To use Disaggregation

## Intermediate API
- Must be authenticated via Auth0
- Must have admin permission via Auth0
- One exception, `get_auth0_user_key_info()` this endpoint is for getting a user's information so whoever with Auth0 token will hit this endpoint right after they logged into the application.

## Project API
- May/May not be authenticated via Auth0
- If a user is authenticated via Auth0, they can access both Private and Public projects.
    - Check if the requested `project_id` is in the list of projects before forwarding it to the Project API.
    - The list of projects are
        - All Public projects
        - All Private projects that a user has permission with
- If a user is not authenticated via Auth0, they can only access Public projects.
    - Check if the requested `project_id` is in the list of public projects before forwarding it to the Project API.

GMHazard_Updated_Authentication_structure drawio (1)

### DIFF
--- a/apis/intermediate_api/intermediate_api/api/core_api.py
+++ b/apis/intermediate_api/intermediate_api/api/core_api.py
@@ -29,13 +29,14 @@ def get_ensemble_ids(is_authenticated):
         return utils.proxy_to_api(
             request, const.ENSEMBLE_IDS_ENDPOINT, "GET", CORE_API_BASE, CORE_API_TOKEN,
         )
-    raise auth0.AuthError(
-        {
-            "code": "Unauthorized",
-            "description": "You don't have access to this resource",
-        },
-        const.NO_ACCESS_RIGHT_CODE,
-    )
+    else:
+        raise auth0.AuthError(
+            {
+                "code": "Unauthorized",
+                "description": "You don't have access to this resource",
+            },
+            const.NO_ACCESS_RIGHT_CODE,
+        )
 
 
 @app.route(const.CORE_API_IMS_ENDPOINT, methods=["GET"])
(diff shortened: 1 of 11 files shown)


apis/intermediate_api/intermediate_api/api/core_api.py
```diff
@@ -23,68 +23,126 @@
...
 def get_ensemble_ids(is_authenticated):
-    return utils.proxy_to_api(
-        request, const.ENSEMBLE_IDS_ENDPOINT, "GET", CORE_API_BASE, CORE_API_TOKEN,
-    )
+    if is_authenticated and auth0.requires_permission("hazard"):
+        return utils.proxy_to_api(
+            request, const.ENSEMBLE_IDS_ENDPOINT, "GET", CORE_API_BASE, CORE_API_TOKEN,
+        )
+    else:
+        raise auth0.AuthError(
```
claudio525, line 33: Can probably make this exception a constant? Also you technically don't need the else

apis/intermediate_api/intermediate_api/api/project_api.py
```diff
@@ -6,7 +6,35 @@
 import intermediate_api.decorators as decorators
 import intermediate_api.auth0 as auth0
 import intermediate_api.constants as const
-import intermediate_api.api.intermediate_api as intermediate_api
+
+
+def has_project_permission(is_authenticated):
```
claudio525, line 11: Thoughts on making this a decorator?

claudio525 (commented):
Looks good!

apis/intermediate_api/intermediate_api/api/project_api.py
SnowNooDLe: Actually, yes, I think this can be done via decorator as by the time the API gets hit, `@app.route()` is the very first one which includes `request`, I originally thought of this as a decorator but because of that `request` variable, I was trying to do it within each endpoint.

apis/intermediate_api/intermediate_api/api/core_api.py
SnowNooDLe: Man, we can even have them as a constant? will check that out.

> Also you technically don't need the else
Yeah as if it doesn't meet if statement condition, will hit this `raise` regardless but added purely for readability :joy: 
It looked slightly uncomfortable with a different indentation...

SnowNooDLe:
Because of the `get_available_projects()["ids"]` usage within the permission checker, I decided to put this decorator within project_api.py instead of decorators.py as decorators need to import project_api to call this function but then, will occur circular import.

claudio525 (approved):

PR #34, opened 2021-11-22
Update Project API's Authentication structure

SnowNooDLe:
# Update Project API's Authentication structure

Basically a split version of #30.
Once three different PRs go through, the `remove_a_few_items` branch should be identical to `update_auth_structure`.

## Project API
- May/May not be authenticated via Auth0
- If a user is authenticated via Auth0, they can access both Private and Public projects.
    - Check if the requested `project_id` is in the list of projects before forwarding it to the Project API.
    - The list of projects are
        - All Public projects
        - All Private projects that a user has permission with
- If a user is not authenticated via Auth0, they can only access Public projects.
    - Check if the requested `project_id` is in the list of public projects before forwarding it to the Project API.

GMHazard_Updated_Authentication_structure drawio (1)

### DIFF
--- a/apis/intermediate_api/intermediate_api/api/project_api.py
+++ b/apis/intermediate_api/intermediate_api/api/project_api.py
@@ -1,3 +1,5 @@
+from functools import wraps
+
 from flask import jsonify, request, Response
 
 from intermediate_api import app, PROJECT_API_BASE, PROJECT_API_TOKEN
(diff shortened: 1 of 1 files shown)


claudio525 (approved):

apis/intermediate_api/intermediate_api/api/project_api.py
```diff
@@ -6,287 +8,403 @@
...
+    Those with authentication, check both private and public projects.
+    Those without authentication, check public projects only.
+    """
+
+    @wraps(f)
+    def decorated(*args, **kwargs):
+        is_authenticated = kwargs["is_authenticated"]
+        project_id = request.args.get("project_id")
+        is_authorized = False
+        if project_id is not None:
```
jasonmotha, line 24: This function's conditionals is a little messy. I think the most simple is to check for project_id is None and return that. (this way we can remove an indentation level.

Also, is the final return ever reached?

jasonmotha (changes requested):

joelridden (approved):
Just probably deal with Jasons comment etc, other than that I'm happy.

apis/intermediate_api/intermediate_api/api/project_api.py
SnowNooDLe: The only reason I chose to have this condition `if project_id is not None:` is because the main task must have an argument called `project_id` so you see the main task first instead of seeing return something if project id is none.

Good catch on checking the final return, from the frontend point of view, no, you will never(let's say 99% instead of 100%) reach this point as the frontend will stop you from sending any requests if the project id is not selected.

The final return check is more for those who try to send requests manually(without using the frontend app) but does not provide project_id.

In that case, response is something like
#### Without the final return
```html
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>500 Internal Server Error</title>
<h1>Internal Server Error</h1>
<p>The server encountered an internal error and was unable to complete your request. Either the server is overloaded or
	there is an error in the application.</p>
```
#### With the final return
(This is not the current return, it's a `raise` but thought this would be more useful to give users/developers more useful warnings instead of saying the second code block which is not authorized.

```json
// with auth0.AuthError
{
    "code": "Invalid Project ID",
    "description": "Please check your Project ID"
}
```
```json
// return f(*args, **kwargs, is_authorized=is_authorized)
{
    "code": "Unauthorized",
    "description": "You don't have access to this resource"
}
```

SnowNooDLe: I was a little confused with
if project_id is None return that but managed it :joy:

joelridden (approved):

jasonmotha (approved):

PR #36, opened 2021-11-25
GMS download no longer needs a token as a parameter

SnowNooDLe:
# GMS download no longer needs a token as a parameter

- Removed token from a parameter(Both Core and Project APIs) to keep the consistency with other endpoints.
- Updated the package version for all APIs - minor version upgrade.

## TODO
1. Do symlink current projects directory to the new directory along with the new project API's package version
2. Update and deploy the latest version of APIs - Both Core and Project
3. Deploy the frontend on 1p


apis/intermediate_api/setup.py
```diff
@@ -2,7 +2,7 @@
 
 setup(
     name="intermediate_api",
-    version="20.8.2",
+    version="20.8.3",
```
joelridden, line 5: Is it also worth to change the core_api version as the core_api endpoints have changed too?

joelridden (commented):
I think the project API also has this token in download_gms_results?
Maybe also update project_api version number if project api does change.

SnowNooDLe:
> I think the project API also has this token in download_gms_results? Maybe also update project_api version number if project api does change.

Good catch, as we don't have dummy GMS data for the current version of project API, didn't even think of the Project API.

jasonmotha (dismissed):
Should update the version if the API changes but otherwise happy.

apis/intermediate_api/setup.py
SnowNooDLe: I think ultimately, it is worth changing it and that 20.8. x, this x position is for minor changes.
However, the intermediate API, I can safely say that it is mainly managed by me with some help from Claudio and Jason whereas Core and Project API, I would put both Claudio and Jason as the main maintainers so not confident enough to change the version number haha.

Also, for the project_api version, I guess we can change it but as soon as we change it, all the endpoints will try to look up the project data under that specific version directory. E.g., if the project_api version is 21.5.1, endpoints look for data under a directory called v21p5p1, so either we need to manually rename the project directory every time we update the project API package version or use another method instead of using package version as a projects directory.
Thoughts? @claudio525 @jasonmotha

jasonmotha: In the future we can make some automated task to run GNZL when we update the API version. In the interim, we shouldn't be doing many changes to the API at the moment - so fine to just deal with it.

claudio525: >Is it also worth to change the core_api version as the core_api endpoints have changed too?

We haven't really been updating these as far as I'm aware, so no good reason to start now (but we should probably start at some point lol)

>Also, for the project_api version, I guess we can change it but as soon as we change it, all the endpoints will try to look up the project data under that specific version directory. E.g., if the project_api version is 21.5.1, endpoints look for data under a directory called v21p5p1, so either we need to manually rename the project directory every time we update the project API package version or use another method instead of using package version as a projects directory.
Thoughts?

I think we should change it so that the directory uses the year-major version, since changes to the project data will always be major. And that way we can do tweaks to the API and just update the minor version number without any problems. But thats probably best done in a separate change. For now, I recon just update the version number and then symlink the GNZL project into the new version directory.

jasonmotha: Agreed with your point about major version. That's a good distinction.

joelridden (approved):

claudio525 (approved):

jasonmotha (approved):

PR #37, opened 2021-11-28
Fix the Core API test

SnowNooDLe:
# Fix the Core API test

- Rename `missingparam` to `missing_param`
- Missing download token
    - Update on status code from 404 to 400
    - Update on content-type from html/text to application/json 
image
- Remove url_extension for Core GMS download
    - TODO: Update the Projects download all then we can eliminate `url_extension` parameter to simplify. 

image

### DIFF
--- a/apis/core_api/core_api/test/test_core_gm_data.py
+++ b/apis/core_api/core_api/test/test_core_gm_data.py
@@ -22,7 +22,7 @@ def test_get_ensemble_ims(config):
     )
 
 
-def test_get_ensemble_ims_missingparam(config):
+def test_get_ensemble_ims_missing_param(config):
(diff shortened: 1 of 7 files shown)


joelridden (commented):
I'm fine with these changes, should change the Project API tests too as it will be a similar change

SnowNooDLe:
> I'm fine with these changes, should change the Project API tests too as it will be a similar change

Yeah, so put TODO in the last bullet point haha.
Just didn't want to modify any extra stuff other than Core API tests.

jasonmotha (approved):

claudio525 (approved):

joelridden (approved):

PR #44, opened 2021-12-09
Fix reset states

SnowNooDLe:
# Fix reset states

## Main changes - Reset the `useState` hook at the right timing
From
```javascript
/*
    Reset when it first renders
    E.g. change tab between Projects to Hazard Analysis
  */
  useEffect(() => {
    setShowImages(false);
    setShowSpinner(false);
    setProjectSiteSelectionGetClick(null);
  }, []);
```
To
```javascript
// Reset some global variables when this component gets unmounted
  useEffect(() => {
    return () => {
      setProjectSiteSelectionGetClick(null);
      setProjectHazardCurveGetClick(null);
      setProjectDisaggGetClick(null);
      setProjectUHSGetClick(null);
      setProjectGMSGetClick(null);
      setProjectScenarioGetClick(null);
    };
  }, []);
```
Code we used to have, theoretically made sense as if a user comes back to Project tab from Hazard Analysis or vice versa, it's an initial render which will trigger that `useEffect(() => {}, [])`. However, because the `useState` hook is asynchronous, `setProjectSiteSelectionGetClick(null);` does not change `projectSiteSelectionGetClick` to null instantly so the app will try to fetch some data but then the app does not have some arguments to go with requests.

This is what happens by switching from Project to Hazard analysis and coming back to Project.
image
and by changing the timing. So, when a user escapes the Projects tab, the following states(global variables) set back to null, so by the time they come back to Projects, nothing will happen.
```javascript
setProjectSiteSelectionGetClick(null);
setProjectHazardCurveGetClick(null);
setProjectDisaggGetClick(null);
setProjectUHSGetClick(null);
setProjectGMSGetClick(null);
setProjectScenarioGetClick(null);
```

## Reset `useStates` when user clicks a get button
Instead of checking multiple variables for the `useEffect` hook, check only one variable.
For instance, the following code gets executed when one of the values(projected, projectVS30, projectLocation) gets changed. However, those are all global variables and they only get changed when a user clicks a get button from the Site Selection tab.
Also, by clicking the get button from the Site Selection, all the values get changed and it causes the app to trigger this useEffect three times.
```javascript
useEffect(() => {
    setShowSpinnerHazard(false);
    setShowPlotHazard(false);
    setProjectHazardCurveGetClick(null);
    setProjectSelectedIM(null);
    setProjectSelectedIMPeriod(null);
  }, [projectId, projectVS30, projectLocation]);
```
Hence, watch the `GetClick` value instead and it only execute `useEffect` hook once a user clicks the Get button from the Site Selection and only if it is not a `null`.
```javascript
// Reset tabs if users click Get button from Site Selection
  useEffect(() => {
    if (projectSiteSelectionGetClick !== null) {
      setShowSpinnerHazard(false);
      setShowPlotHazard(false);
      setProjectHazardCurveGetClick(null);
      setProjectSelectedIM(null);
      setProjectSelectedIMPeriod(null);
    }
  }, [projectSiteSelectionGetClick]);
```

## Minor changes
- Change the order of `useEffect`, any `useEffect` to reset variables will now be at the top of the component. (Still below any `useState`)
- Changed a way of checking the size of the array from `a !== []` to `a.length !== 0` as with JS, `[] !== []` will always be true.
    - Speaking of checking the size of the array, we can perhaps think of replacing `a.length !== 0` to `a.length > 0` as the length of the array cannot be negative and we want them to be not empty, which is always greater than 0 but I haven't pushed that far within this PR.
- When a user clicks a Get within the Scenario tab, the selected scenarios will be removed.

### DIFF
--- a/frontend/CHANGELOG.md
+++ b/frontend/CHANGELOG.md
@@ -1,3 +1,14 @@
+## December 9, 2021
+
+### Reset useState hooks at the right timing - ([PR #4](https://github.com/ucgmsim/gmhazard/pull/44))
+
+- Main changes
+  - Reset the `useState` hook at the right timing
+  - Reset any useStates when the user clicks a get button from the Site Selection tab
+- Minor changes
+  - Change the order of useEffect, any useEffect to reset variables will now be at the top of the component. (Still below any `useState`)
+  - Changed the way of checking the size of the array from a !== [] to a.length !== 0 as with JS, [] !== [] will always be true.
+
 ## December 1, 2021
 
 ### Deploying GMHazard web app to subdirectory - ([PR #39](https://github.com/ucgmsim/gmhazard/pull/39))
--- a/frontend/CHANGELOG.md
+++ b/frontend/CHANGELOG.md
@@ -2,12 +2,8 @@
 
 ### Reset useState hooks at the right timing - ([PR #4](https://github.com/ucgmsim/gmhazard/pull/44))
 
-- Main changes
-  - Reset the `useState` hook at the right timing
-  - Reset any useStates when the user clicks a get button from the Site Selection tab
-- Minor changes
-  - Change the order of useEffect, any useEffect to reset variables will now be at the top of the component. (Still below any `useState`)
-  - Changed the way of checking the size of the array from a !== [] to a.length !== 0 as with JS, [] !== [] will always be true.
+- Reset the states when the Projects view component gets unmounted to prevent sending a request.
+  - This fixes the issue that the app may get crashed when they access to Projects page after visiting Home or Framework Documents.
 
 ## December 1, 2021
 
--- a/frontend/src/components/Project/SeismicHazard/HazardViewerDisaggregation.js
+++ b/frontend/src/components/Project/SeismicHazard/HazardViewerDisaggregation.js
@@ -39,6 +39,7 @@ const HazadViewerDisaggregation = () => {
     projectSelectedIMComponent,
     projectSelectedDisagRP,
     setProjectSelectedDisagRP,
+    projectSiteSelectionGetClick,
   } = useContext(GlobalContext);
 
   // For data fetcher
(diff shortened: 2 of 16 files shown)


frontend/src/components/Project/SeismicHazard/HazardViewerDisaggregation.js
```diff
@@ -70,6 +71,24 @@ const HazadViewerDisaggregation = () => {
     src: null,
   });
 
+  // Reset tabs if users change project id, project vs30, project location or project im
```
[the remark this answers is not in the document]
SnowNooDLe: Glad I changed everywhere else...

frontend/CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+## December 9, 2021
+
+### Reset useState hooks at the right timing - ([PR #4](https://github.com/ucgmsim/gmhazard/pull/44))
+
+- Main changes
+  - Reset the `useState` hook at the right timing
+  - Reset any useStates when the user clicks a get button from the Site Selection tab
+- Minor changes
+  - Change the order of useEffect, any useEffect to reset variables will now be at the top of the component. (Still below any `useState`)
+  - Changed the way of checking the size of the array from a !== [] to a.length !== 0 as with JS, [] !== [] will always be true.
```
jasonmotha, line 10: Generally, in changelogs it should be user-facing changes listed rather than code notes. So how each change impacts the user experiienc. The second comment is almost there. Thoughts?

jasonmotha (approved):

frontend/CHANGELOG.md
SnowNooDLe: Fair call, I believe I got the exact feedback from Claudio once back then haha.
I will try to write something in user-facing rather than ourselves.

joelridden (approved):

jasonmotha (approved):

claudio525 (approved):

PR #50, opened 2022-01-11
Spatial Hazard Refactor

joelridden:
### **Spatial Hazard Correlate IM's Refactor**

Adjusted correlate IM's functions
Added rupture lookup to calc_emp_flt for generetd db's so they can be used for spatial hazard


calculation/spatial_hazard/spatial_hazard/correlate_ims.py
```diff
@@ -78,150 +77,85 @@ def calculate_distance_matrix(
...
-        A3 += I * (-mineig * k ** 2 + spacing)
-        k += 1
+    # Calculate random between event residual value per realisation and multiply by between event sigma
+    between_event = (
+        np.random.normal(0.0, 1.0, size=N)[:, None] * between_event_std[None, :]
+    )
 
-    return A3
+    # Calculate random within event residual value per site and per realisation
+    # Calculate random within event residual value per site and per realisation
```
SnowNooDLe, line 152: Duplicated comment! haha

SnowNooDLe (approved):
Sweet, I could only find one nitpicking comment :+1:

calculation/gmhazard_calc/gmhazard_calc/dbs/IMDB.py
```diff
@@ -330,7 +330,7 @@ def rupture_names(self) -> Union[None, np.ndarray]:
 
     @check_open
     def im_data(
-        self, station: str, im: Optional[Union[List[IM], IM]] = None, inter_intra: bool = False,
+        self, station: str, im: Optional[Union[List[IM], IM]] = None, incl_inter_intra: bool = False,
```
claudio525, line 333: I recon `incl_with_between_sigma` or `incl_rand_effects_sigma` might be better? Not a fan of `inter` and `intra`

calculation/gmhazard_calc/gmhazard_calc/dbs/IMDB.py
```diff
@@ -341,7 +341,7 @@ def im_data(
         im: IM or list[IM], optional
             IM(s) of interest
             if this is unspecified then it is equivalent to setting all IMs
-        inter_intra: bool
+        incl_inter_intra: bool
             Boolean flag to determine to either extract normal mu and sigma or mu, sigma_inter and sigma_intra
```
claudio525, line 345: Not yours, but still worth fixing. Should be 
`..... extract mu and total standard deviation or mu, between-event and within-event standard deviation`

calculation/gmhazard_calc/gmhazard_calc/dbs/SiteSourceDB.py
```diff
@@ -83,6 +83,21 @@ def has_station_data(self, station_name):
 
         return key in self._keys_cache
 
+    @check_open
+    def has_station_directivity_data(self, station_name):
```
claudio525, line 87: How come this is in this PR? Doesn't really have anything to do with spatial hazard?

calculation/spatial_hazard/spatial_hazard/correlate_ims.py
```diff
@@ -5,65 +5,64 @@
...
+    imdb_ffps: Sequence[str]
+        List of imdb full file paths to load
+    stations: Sequence[str]
+        List of station names to grab data for from the imdbs
+    im: IM
+        IM Object to extract values from the imdb
+    fault: String
+        Fault name to extract data from the imdb
+    """
+    # Obtain the given IMDB with the correct fault information
```
claudio525, line 29: No `given`?
Also I would also make clear here, that the assumption we are making is that for a given fault there is only a single IMDB that will contain the required data. Due to the fact that spatial hazard does not support Ensembles, i.e. only works for a single GMM.

calculation/spatial_hazard/spatial_hazard/correlate_ims.py
```diff
@@ -78,150 +77,85 @@ def calculate_distance_matrix(
...
+        correlation = sha_calc.spatial_hazard.loth_baker_model.get_correlations(
             str(selected_im),
             str(selected_im),
-            distance_matrix.loc[stations, station_i].values,
+            distance_matrix.loc[stations, station].values,
         )
         assert np.all(
             correlation >= 0.0
-        ), "Correlation should be positive or 0. Error with {}, {}".format(i, station_i)
+        ), f"Correlation should be positive or 0. Error with {i}, {station}"
```
jasonmotha, line 105: Does this error message make sense? Error with `i` where I is a station index?

calculation/spatial_hazard/spatial_hazard/correlate_ims.py
```diff
@@ -78,150 +77,85 @@ def calculate_distance_matrix(
...
         R[i, :] = R[:, i] = correlation
 
-    R = pd.DataFrame(index=stations, columns=stations, data=R)
-    return R
+    return pd.DataFrame(index=stations, columns=stations, data=R)
 
 
 def create_random_fields(
```
claudio525, line 111: I think this can be renamed to something `gen_IM_values` or similar, makes it a bit clearer what it does imo

calculation/spatial_hazard/spatial_hazard/correlate_ims.py
```diff
@@ -78,150 +77,85 @@ def calculate_distance_matrix(
...
     N: int,
     R: pd.DataFrame,
-    stations: Sequence[str],
+    n_stations: int,
```
claudio525, line 114: I don't think this parameter is needed, since this should always be the same as the first dimension of `emp_df`?

tools/gmhazard_scripts/db_creation/empirical_db/common.py
```diff
@@ -170,6 +172,8 @@ def write_data_and_close(
             ims=np.asarray(im_list, dtype=str),
         )
         imdb_dict[imdb_keys].close()
+        if rupture_lookup:
+            imdb_dict[imdb_keys].add_rupture_lookup(imdb_dict[imdb_keys].db_ffp, 1)
```
jasonmotha, line 176: Don't think this should be after the close?

This is potentially dangerous as it shouldn't run for ds (it will quickly run out of RAM I think - if not take forever to complete). We should add some checking to this method to only do it if it is a fault db and error otherwise.

jasonmotha: Oh, the method does open it? It's not clear from a cursory glance. If it's a larger db it probably needs more than 1 processor too. But fault normally only runs in a single process though.

tools/gmhazard_scripts/db_creation/empirical_db/calc_emp_flt.py
```diff
@@ -74,6 +75,7 @@ def calculate_flt(
                     tect_type_model_dict_ffp,
                     max_dist,
                     keep_sigma_components=keep_sigma,
+                    use_directivity=distance_store.has_station_directivity_data(site.name)
```
claudio525, line 78: Something from directivity?

tools/gmhazard_scripts/db_creation/empirical_db/common.py
claudio525, line 176: As far as I understand this is called by the MPI master process at the end, in which case there is no reason to not set `n_procs` to all available process?

tools/gmhazard_scripts/db_creation/empirical_db/common.py
```diff
@@ -309,11 +313,12 @@ def calculate_emp_site(
         distance_df, left_on="fault_name", right_on="fault_name"
     )
 
-    directivity_df = fault_df.merge(
-        distance_store.station_directivity_data(station_name),
-        left_on="fault_name",
-        right_index=True,
-    )
+    if use_directivity:
```
claudio525, line 316: More directivity?

tools/gmhazard_scripts/db_creation/empirical_db/common.py
claudio525: Didn't see Jason's comment, so probably ignore what I said above haha.

calculation/gmhazard_calc/gmhazard_calc/dbs/IMDB.py
```diff
@@ -330,7 +330,7 @@ def rupture_names(self) -> Union[None, np.ndarray]:
 
     @check_open
     def im_data(
-        self, station: str, im: Optional[Union[List[IM], IM]] = None, inter_intra: bool = False,
+        self, station: str, im: Optional[Union[List[IM], IM]] = None, incl_with_between_sigma: bool = False,
```
SnowNooDLe, line 333: one nitpicking, if we decide to change the name from `inter_intra` to something more meaningful,
Shall we use `incl_within_between_sigma`? 
It feels slightly odd to me seeing with and between in the variable and based on the docstring, it says `between-event` and `within-event` instead of `with-event`, so when the variable got longer, might as well add extra 2 characters? :thinking: 
However, this is my personal opinion, so up to you to keep or change it :+1:

SnowNooDLe (approved):
Another one nitpicking opinion, up to you to decide :+1:

SnowNooDLe (dismissed):

tools/gmhazard_scripts/db_creation/empirical_db/common.py
claudio525: @joelridden Agree with @jasonmotha on this one, probably good to add assert to ensure this only runs for fault IMDB generation

calculation/spatial_hazard/spatial_hazard/correlate_ims.py
claudio525: @joelridden Thoughts on adding extra comment regarding assumptions made?

tools/gmhazard_scripts/db_creation/empirical_db/common.py
joelridden: @claudio525 Not sure if an assert is necessary as the rupture lookup variable is default to false and can currently only be set from the calc_emp_flt script, therefore it won't actually run for DS unless you manually grab a DS db and run that function.
At that point though I feel like its the users fault if they are trying to run this function for DS? Maybe a comment about it saying a warning for running this on any DS db?

claudio525: Ah yeah, that sounds all good

SnowNooDLe (approved):

claudio525 (approved):

jasonmotha (approved):

PR #51, opened 2022-01-12
directivity data check

joelridden:
### **Site Source Directivity data fix**

Bug for site source db's with no directivity data not being able to calculate emperical fault dbs.
Now added a check for directivity data before trying to merge the dataframes in calc_emp_site function.


jasonmotha (approved):

claudio525 (approved):

SnowNooDLe (approved):
It feels like a deja vu, I've seen these before :thinking:

PR #53, opened 2022-01-20
Mp directivity

claudio525:


calculation/gmhazard_calc/gmhazard_calc/directivity/directivity.py
```diff
@@ -157,40 +159,63 @@ def compute_fault_directivity(
...
-            plane_index,
-            sites,
-            nominal_strike,
-            nominal_strike2,
-            mag,
-            rake,
-            periods,
-        )
+            fdi.append(cur_fdi)
+            phired.append(cur_fdi)
```
joelridden: I'm assuming that this is for performance using a list instead of numpy index as RAM isn't an issue anymore?

calculation/gmhazard_calc/gmhazard_calc/directivity/directivity.py
```diff
@@ -157,40 +159,63 @@ def compute_fault_directivity(
...
+                        rake,
+                        periods,
+                    )
+                    for (ix, cur_plane) in enumerate(hypo_planes)
+                ],
+            )
+
+        # Combine results
+        fdi = [cur_result[0] for cur_result in results]
+        phired = [cur_result[1][0] for cur_result in results]
```
joelridden, line 202: Could this be done like this?
`fdi, phired = list(zip(*results))`
Just feels cleaner to me

calculation/gmhazard_calc/gmhazard_calc/directivity/directivity.py
```diff
@@ -157,40 +159,63 @@ def compute_fault_directivity(
...
-            plane_index,
-            sites,
-            nominal_strike,
-            nominal_strike2,
-            mag,
-            rake,
-            periods,
-        )
+            fdi.append(cur_fdi)
+            phired.append(cur_phi_red)
```
joelridden, line 178: Previous comment got removed by refresh haha
Is the reason for lists instead of numpy indexing to improve performace as RAM isnt an issue anymore?

joelridden (commented):
Just a small thing

calculation/gmhazard_calc/gmhazard_calc/directivity/directivity.py
SnowNooDLe: I think it can't be? if the period is `cur_result[1]` then might work instead of `cur_result[1]` 🤔

tools/gmhazard_scripts/db_creation/empirical_db/calc_distances_flt.py
```diff
@@ -1,4 +1,6 @@
 import argparse
+import time
+
 import numpy as np
 import os
```
SnowNooDLe, line 5: while there was a change in line 2~3, might as well we move `import os` to the top 👍

SnowNooDLe (dismissed):

calculation/gmhazard_calc/gmhazard_calc/directivity/directivity.py
```diff
@@ -157,40 +159,63 @@ def compute_fault_directivity(
...
+            )
+
+        # Combine results
+        fdi = [cur_result[0] for cur_result in results]
+        phired = [cur_result[1][0] for cur_result in results]
 
-        fdi_array[index] = fdi
-        phired_array[index] = phi_red
+    fdi = np.stack(fdi, axis=0)
+    phired = np.stack(phired, axis=0)
```
jasonmotha, line 205: I'd prefer this with the stack done on the one line - I think python can optimise the nested call a little.

calculation/gmhazard_calc/gmhazard_calc/directivity/directivity.py
```diff
@@ -157,40 +159,63 @@ def compute_fault_directivity(
...
 
     if n_hypo_data.method == constants.HypoMethod.UNIFORM_GRID:
         # Only apply weights if method type is uniform grid
-        fdi_average = weights[:, None, None] * fdi_array
-        phired_average = weights[:, None, None] * phired_array
+        fdi_average = weights[:, None, None] * fdi
+        phired_average = weights[:, None, None] * phired
```
jasonmotha, line 210: again, could revise to one-liners.

tools/gmhazard_scripts/db_creation/empirical_db/calc_distances_flt.py
```diff
@@ -142,16 +152,22 @@ def load_args():
...
     )
     parser.add_argument(
         "--gcmt_ffp", type=str, help="Path to the GCMT csv file", default=None
     )
     parser.add_argument(
-        "station_file",
-        help="List of stations for a specific domain. Source to site distances "
-        "will be calculated for all stations in the station_file.",
+        "--n_procs",
+        help="Number of processes to use for the directivity calculation",
```
jasonmotha, line 168: Could mention that the numba uses all available - but not a big deal.

calculation/gmhazard_calc/gmhazard_calc/directivity/directivity.py
claudio525: There isn't any real RAM difference between this one and what was there previously. Also (not sure the context we talked about this last) the problem in general isn't RAM limitations, but unneeded memory copies (which you get when using append on a numpy array/pandas dataframe).
The change I have made is very similar to what was there, there is no increase in memory copies (I'm happy to explain, but have a think about it yourself first).

calculation/gmhazard_calc/gmhazard_calc/directivity/directivity.py
claudio525: Hmh maybe for `fdi` something like `fdi = list(list(zip(results))[0])` would work, but it
a) wouldn't work for `phired`
b) I don't think this is any cleaner, much harder to read. 
One of the reasons why I list list comprehensions are pretty simple to read (once you are know the syntax) compared to things like `apply` etc.

calculation/gmhazard_calc/gmhazard_calc/directivity/directivity.py
claudio525: I'm not  100% sure what you mean. Do you mean making it a one-liner like this `fdi, phired = np.stack(fdi, axis=0), np.stack(phired, axis=0)`?

jasonmotha: I mean like `fdi = np.stack([cur_result[0] for cur_result in results], axis=0)`

or convert the whole results "mess" into a numpy array and slice it.

joelridden (approved):

SnowNooDLe (dismissed):
Just realized that the file `calc_distances_flt.py`(which is not related to the directivity I assume) is within this PR but it is also in another PR, I will have a look more deeply in the MP calc_emp_flt PR.

calculation/gmhazard_calc/gmhazard_calc/directivity/directivity.py
claudio525: Then I would have to indent it and do a separate stack for the for loop. 
And I don't think you can just turn into a numpy array and slice it, since its a Tuple that contains another tuple.

joelridden (approved):

calculation/gmhazard_calc/gmhazard_calc/directivity/directivity.py
jasonmotha: Oh yep. I didn't notice the indentation.

calculation/gmhazard_calc/gmhazard_calc/directivity/directivity.py
jasonmotha: You could do it here though?

jasonmotha (approved):

PR #55, opened 2022-02-03
changed the z writer to use the csv format too

jasonmotha:
Writes as a csv for the z values.

tested with this reader to make sure it's the same format.
https://github.com/ucgmsim/qcore/pull/263

### DIFF
--- a/tools/project_gen/project_gen/project.py
+++ b/tools/project_gen/project_gen/project.py
@@ -100,10 +100,7 @@ def create_project(
     erf_dir = ERF_DIR if erf_dir is None else erf_dir
 
     try:
-        projects_base_dir, scripts_dir = (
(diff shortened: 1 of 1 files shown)


tools/project_gen/project_gen/project.py
```diff
@@ -277,6 +287,7 @@ def write_file_data(station_detail_indexes: List, station_details: List, ffp: Pa
                     loc_data["lon"],
                     np.nan if z1p0 is None else z1p0,
                     np.nan if z2p5 is None else z2p5,
+                    0,
```
joelridden, line 290: This 0 feels very magical, I'm not exactly sure what this is?

joelridden (approved):

claudio525 (approved):

PR #56, opened 2022-03-07
Distribution Computation Time Fix

joelridden:
Fixing time issue with creating distributions for hypocentre locations taking from 1.5 to 3 seconds for each fault.
Reduced now to 0.004 to 0.007 seconds.

Fixed by removing the custom scipy distributions as they were needing to convert a pdf to a ppf which was slow.
Instead now truncate by multilpying the latin hypercube values by the cdf of the max depth (1) and then lookup the inverse cdf for getting the results.

Below is a quick test with the gamma distribution to ensure the same results were computed between these different approaches.
hypo_dist
hypo_dist3


jasonmotha (approved):

SnowNooDLe (approved):
When I heard the pdf this morning, I thought it was a document pdf not the probability density function :exploding_head:

claudio525 (approved):

PR #57, opened 2022-03-09
Moving GC2 out of hypocentre loop

joelridden:
Now only needs to calculate GC2 once for each fault for directivity.
Also combined nominal strike with site coords as they are just site locations and are extracted out after GC2 is calculated.


SnowNooDLe (approved):
I think they are good.

I almost left a comment to check on lines 180~182 but realized 182 has colon symbol when 180 doesn't, I originally saw `ry_start` and `ry` were same but `_compute_directivitiy_effect` was saying they were two different data types :thinking:

jasonmotha (approved):

claudio525 (approved):

PR #58, opened 2022-03-21
HOTFIX: calc_emp_flt module - the parse arg

SnowNooDLe:
# HOTFIX: calc_emp_flt module - the parse arg

Got the following error message
```python
TypeError: __init__() got an unexpected keyword argument 'type'
```
I don't think that type is valid for boolean type and add action to `rupture_lookup` as well.

Hence, I used one of the [options](https://www.pythonpool.com/python-argparse-boolean/), the second one.

### DIFF
--- a/calculation/gmhazard_calc/gmhazard_calc/dbs/SiteSourceDB.py
+++ b/calculation/gmhazard_calc/gmhazard_calc/dbs/SiteSourceDB.py
@@ -7,6 +7,10 @@ from gmhazard_calc import constants as const
 from .BaseDB import BaseDB, check_open
 
 
+STATION_DIRECTIVITY = "/directivity/station"
+STATION_DISTANCE = "/distances/station"
+
+
 class SiteSourceDB(BaseDB):
     """Class that handles retrieving/writing data with a SiteSourceDB"""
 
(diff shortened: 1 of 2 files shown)


claudio525 (dismissed):
I'm sure I have always done that, but apparently not lol. What a pleb..

jasonmotha (dismissed):

calculation/gmhazard_calc/gmhazard_calc/dbs/SiteSourceDB.py
```diff
@@ -51,7 +51,7 @@ def stored_stations(self):
             self._keys_cache = set(self._db.keys())
 
         return [
-            cur_key.split("_")[-1]
```
jasonmotha, line 54: I'd put a comment here to explain what it's doing / the structure of a key. "removes the '/distance/station_' at the start of the key to retrieve the station name"

It's out of your scope but I'll put "/distance" into a constant as it's starting to be referred to multiple times in this code.

jasonmotha (changes requested):

calculation/gmhazard_calc/gmhazard_calc/dbs/SiteSourceDB.py
SnowNooDLe: @claudio525 In case you missed it :smiley:

claudio525: @SnowNooDLe Can you add that comment and add that constant pls

joelridden (approved):

claudio525 (approved):

jasonmotha (approved):

PR #59, opened 2022-03-29
New calc_emp_ds

SnowNooDLe:
# New calc_emp_ds

## We originally planned to do something like
Image from iOS (4)

## However,
But by looking at the existing script, I thought it would be simple to keep the current workflow.
For instance, 
1. function `write_result_to_db()` takes an arguments with `station_name` to store `im_data` per station.
2. common module's function, `get_work()` returns the number of stations that are not computed yet as a checkpoint.

### Remove MPI & use openquake_vectorized_wrapper
By removing MPI and adopting vectorized calculation, we don't have to go through every single rupture from `matching_df`, they can be calculated at once as our `vectorzied_oq_wrapper` is designed to work one multiple ruptures but with a single site.
(Any code below the comment, `# Adding missing columns` will make this `matching_df` contain three different types for OQ, such as Site, Distance and Rupture which is the form of the data frame we need to pass to our `vectorized_oq_wrapper`)

Also, because we for loop over `tect_type`, I chose the function `__new_process_rupture` to update one of its params, `fault_im_result_dict` instead of creating a new one every single run(Like the one with the previous version) as we do not need them anymore.

NEW
```python
fault_im_result_dict = {key: [] for key in imdb_dict.keys()}
    for tect_type in matching_df["tect_type"].unique():
        __new_process_rupture(
            matching_df,
            rupture_df,
            im_types,
            tect_type,
            tect_type_model_dict,
            directivity_df if use_directivity else None,
            psa_periods,
            keep_sigma_components,
            fault_im_result_dict,
            use_directivity,
        )
```

PREVIOUS
```python
results = p.starmap(
                __process_rupture,
                [
                    (
                        rupture,
                        site,
                        im_types,
                        tect_type_model_dict,
                        directivity_df.iloc[index] if use_directivity else None,
                        psa_periods,
                        keep_sigma_components,
                        {key: {} for key in imdb_dict.keys()},
                        use_directivity,
                    )
                    for index, rupture in matching_df.iterrows()
                ],
```

## Result
Six different combinations of Model and tect_type got created.
based on total 72 stations (24 stations * 3 different Vs30 values) finished in about 7.2 mins
image

### Comparison between OLD and NEW PGA/PGA_mean
image

Interestingly indexes are all equal but some values are not, 9 values out of 38092 are not identical.
image

### DIFF
--- a/tools/gmhazard_scripts/db_creation/empirical_db/calc_emp_ds.py
+++ b/tools/gmhazard_scripts/db_creation/empirical_db/calc_emp_ds.py
@@ -122,6 +122,8 @@ def calculate_ds(
             for i, n_rows, site, n_mp_proc in iter(
                 lambda: comm.sendrecv(value, dest=master), StopIteration
             ):
+                print("IN CALC_EMP_DS")
+                print(site)
                 comm_time = time.perf_counter() - s_time
                 wait_time += comm_time
                 if distance_store.has_station_data(site.name):
--- a/tools/gmhazard_scripts/db_creation/empirical_db/calc_emp_ds.py
+++ b/tools/gmhazard_scripts/db_creation/empirical_db/calc_emp_ds.py
@@ -122,8 +122,6 @@ def calculate_ds(
             for i, n_rows, site, n_mp_proc in iter(
                 lambda: comm.sendrecv(value, dest=master), StopIteration
             ):
-                print("IN CALC_EMP_DS")
-                print(site)
                 comm_time = time.perf_counter() - s_time
                 wait_time += comm_time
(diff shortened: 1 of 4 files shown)


jasonmotha:
If you are using isclose to check for float comparison how different are the values?

SnowNooDLe:
> If you are using isclose to check for float comparison how different are the values?

Using np.isclose() returns all True
```cmd
In [130]: np.isclose(new.values, old.values).all()
Out[130]: True
```

SnowNooDLe:
Got some feedback and chat in person, close this PR for now

PR #60, opened 2022-03-30
New calc emp ds

SnowNooDLe:
# New calc emp ds

The structure is based on our discussion.
For 
- IM
    - GMM
        - Site
        - wirte_im_data after each site calculation

### Update on the 11th of April, 2022

1. Adopt feedback
   - Create a helper function for `wirte_metadata_and_close()` called `write_metadata()` which is used by new_calc_emp_ds and `write_metadata_and_close()`
   - Get valid sites using `np.isin()` to provide better performance when there are many stations.
   - Using `create_parametric_db_name` to create a DB filename, to keep the consistency.
   - Print those models that are not supported yet.
2. Changed the labels - same with the `calc_emp_ds` so this way, we can use `im_data(station, im)` to print `mu` and `sigma`
    - PGA_mean -> PGA
    - PGA_std_Total -> PGA_sigma 

### Need to check where it says False - Except BSSA_14 for now
image (10)

### Update on the 8th of April, 2022

1. Updates based on the feedback
4. Changed a way of writing to DB
    - Instead of creating a whole possible DBs at the beginning, we now create just before we start calculating, that way we only get DBs we actually need, no need to keep opening/closing the DB as we did once every site computation is done.
image

### Update on the 4th of April, 2022
1. If Vs30 is not provided, will use the default one from EE's classdef module
5. Z1.0 and Z2.5 are estimated by the given Vs30 if not provided.
6. Combined two functions into one to get a rupture_context_df
7. Added type hints
8. Minor tidy up

#### IMDB's `add_im_data()`
Added the else: cur_df = im_df lines as if the current IMDB does not have any station data, then cur_df is None.
Then instead of putting the given im_df, we are trying to write_im_data with cur_df which is None.
```python
        try:
            cur_df = self._db[path]
        except KeyError:
            pass

        if cur_df is not None:
            # Check that the number of ruptures match
            assert cur_df.shape[0] == im_df.shape[0]

            # Delete the existing one
            self._db.remove(path)

            # Create updated
            cur_df = pd.concat([cur_df, im_df], axis=1)
        else:
            cur_df = im_df
        # Write
        self.write_im_data(station_name, cur_df)
```

#### `max_dist` variable
W used to get this `max_dist` by
```python
# max_rjb == max_dist = common.get_max_dist_zfac_scaled(site)
    if dist_filter_by_mag:
        max_dist = np.minimum(np.interp(matching_df.mag, MAG, DIST), max_rjb)
    else:
        max_dist = max_rjb
```
However, for calc_emp_ds.py, we always passed the `dist_filter_by_mag=True` hence with the newer version of calc_emp_ds.py, will just be like
```python
max_dist = np.minimum(
        np.interp(rupture_df.mag, MAG, DIST),
        common.get_max_dist_zfac_scaled(site_data),
    )
```


tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
```diff
@@ -0,0 +1,253 @@
+import argparse
+import time
+import math
+
+import pandas as pd
+import numpy as np
+
+import common
+import gmhazard_calc as gc
+from empirical.util import empirical_factory, classdef, openquake_wrapper_vectorized
```
claudio525, line 10: Import the module pls

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
```diff
@@ -0,0 +1,253 @@
...
+    Combining the given rupture_context_df(Distance and Rupture) with Site information
+    to create a OQ's RuptureContext like DF to be used with OQ's Vectorized Wrapper
+    """
+    # Adding missing columns
+    # Site Parameters
+    rupture_context_df["vs30"] = site_data.vs30
+    rupture_context_df["vs30measured"] = (
+        site_data.vs30measured if site_data.get("vs30measured") is not None else False
+    )
+    rupture_context_df["z1pt0"] = (
```
claudio525, line 77: These two can be simplified to 
`rupture_context_df["z1pt0"] = site_data.get("z1p0"]`

And then if OQ actually can't handle nan values, use fillna to convert nan to None, i.e.
`rupture_context_df[["z1...", "z2..."]] = `rupture_context_df[["z1...", "z2..."]].fillna(...)`

claudio525 (commented):
Looking good

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
SnowNooDLe: Surprisingly they are modules haha
image

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
```diff
@@ -0,0 +1,253 @@
...
+    max_dist = np.minimum(
+        np.interp(rupture_df.mag, MAG, DIST),
+        common.get_max_dist_zfac_scaled(site_data),
+    )
+    rupture_df = rupture_df[rupture_df["rjb"] < max_dist]
+
+    return polish_rupture_context_df(rupture_df, site_data, tect_type)
+
+
+def polish_rupture_context_df(
```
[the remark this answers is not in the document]
SnowNooDLe: We don't actually need one but felt bad when I saw a massively long one function haha

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
SnowNooDLe: Within the common module, get_work function does something like
```python
site_df = station_df.merge(vs30_df, left_index=True, right_index=True)

    # Combine Z-values and site data (if available)
    site_df = (
        site_df.join(formats.load_z_file(z_file)) if z_file is not None else site_df
    )
```
and while I am commenting on this, I noticed that to run calc_emp_ds, it must have z_file to pass anyway and site_df must have z values. then I can just do something like
`df = df.fillna(0)` instead and  just rename a few columns

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
```diff
@@ -0,0 +1,246 @@
...
+):
+    nhm_data = gc.utils.ds_nhm_to_rup_df(background_sources_ffp)
+    rupture_df = pd.DataFrame(nhm_data["rupture_name"])
+    imdb_dict, _ = common.open_imdbs(
+        model_dict_ffp, output_dir, gc.constants.SourceType.distributed, suffix=suffix,
+    )
+    model_dict = empirical_factory.read_model_dict(model_dict_ffp)
+
+    with gc.dbs.SiteSourceDB(site_source_db_ffp) as distance_store:
+        fault_df, _, site_df, _ = common.get_work(
```
claudio525, line 110: Since neither ds or flt use MPI I think we should rename the `get_work` function and remove the MPI components, i.e. `rank`, `size` (inputs) and `work` (outputs).

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
```diff
@@ -0,0 +1,246 @@
...
+        help="model dictionary to specify which model to use for each tect-type",
+    )
+    parser.add_argument(
+        "--suffix", "-s", help="suffix for the end of the imdb files", default=None,
+    )
+
+    return parser.parse_args()
+
+
+def calculate_emp_ds():
```
claudio525, line 228: Don't think this function is needed, just put that code under `if __name__ == "__main__"` directly?

claudio525 (commented):
Looks good, I assume you have checked that the outputs line up with the old outputs?

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
SnowNooDLe: I think this can be done but it's kinda related to the reason I set `new_calc_emp_ds.py` instead of modifying the `calc_emp_ds`.

we probably need to confirm but as long as we need all the models that the current config has. E.g., K_20, ZA_06, CB_14 and ASK_14. (K_20 and ZA_06 does not support the standard periods, CB_14 and ASK_14, we need to implement width estimation), we should still need to use the `calc_emp_ds`

As the existing `calc_emp_ds` works, we just cannot run with many processes so until we move on to the new_calc_emp_ds, we might have to keep the `get_work` as is for now? otherwise can be done, but I guess it might be better to do it in another PR to simplify `get_work` then apply those changes to two files(ds and flt)

jasonmotha: After this is merged, until we sort out periods and width estimation there still is a use case for the old workflow. But once we sort those issues, we can edit the get_work function and delete the old workflow.

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
jasonmotha: Ahhh, the old Claudio vs Jason style. I prefer the current way so it doesn't chuck a bunch of stuff in the global namespace. But Claudio prefers the other way. And we still haven't came to an agreement. :toilet:

tools/gmhazard_scripts/db_creation/empirical_db/empirical_model_configs/21p10.yaml
```diff
@@ -91,8 +91,6 @@ SUBDUCTION_SLAB:
   PGA:
     geom:
       - ZA_06
-      - BCH_16
-      - A_18  # ZA06 or BCH16
```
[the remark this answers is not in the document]
jasonmotha: Reminder of this comment too.

jasonmotha (changes requested):

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
```diff
@@ -0,0 +1,246 @@
...
+DIST = [125, 150, 175, 200, 250, 300]
+
+
+def get_rupture_context_df(
+    distance_store: gc.dbs.SiteSourceDB,
+    fault_df: pd.DataFrame,
+    site: str,
+    site_data: pd.Series,
+    nhm_data: pd.DataFrame,
+    tect_type: str,
```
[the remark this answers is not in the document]
claudio525: I'm pretty sure we have one in GMHazard

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
```diff
@@ -0,0 +1,246 @@
...
+                GMMs = empirical_factory.determine_all_gmm(
+                    classdef.Fault(tect_type=classdef.TectType[tect_type]),
+                    str(im),
+                    model_dict,
+                )
+                for GMM_idx, (GMM, _) in enumerate(GMMs):
+                    db_type = f"{GMM.name}_{tect_type}"
+                    if GMM.name in ("K_20", "K_20_NZ", "ZA_06"):
+                        continue
+                    print(f"Processing DB Type: {db_type}, {GMM_idx + 1} / {len(GMMs)}")
```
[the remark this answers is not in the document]
SnowNooDLe: I shall probably rename it to `Processing Model: something` instead.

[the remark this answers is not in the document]
jasonmotha: Processing {model} for {tect-type} I reckon?

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
```diff
@@ -101,95 +90,112 @@ def calculate_ds(
...
-                            gmm_calculated_df.loc[
-                                :,
-                                gmm_calculated_df.columns.str.endswith(
-                                    ("mean", "sigma")
+
+                    db_type = f"{GMM.name}_{tect_type}"
+                    suffix = f"_{suffix}" if suffix else ""
+                    # Create a DB if not exists
+                    imdb = gc.dbs.IMDBParametric(
+                        str(Path(output_dir) / f"{db_type}_ds{suffix}.db"),
```
jasonmotha, line 133: I think there's a function in utils called create_db_name?

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
```diff
@@ -101,95 +90,112 @@ def calculate_ds(
...
-        ims,
-        model_dict_ffp,
-    )
-    print("Done")
+                        imdb.write_attributes(
+                            Path(background_sources_ffp).name,
+                            Path(vs30_ffp).name,
+                            ims=np.asarray(im_list, dtype=str),
+                        )
+                        print(f"Writing metadata for Model: {GMM.name} is done.")
```
jasonmotha, line 198: I'd argue to still use the function (just for consistency sake) - probably best to either create a sub_function that writes metadata for one db or to call the existing function with a dummy dictionary for the single db.

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
```diff
@@ -0,0 +1,248 @@
...
+                    f"Processing Tectonic Type: {tect_type}, {tect_idx + 1} / {len(tect_types)}"
+                )
+                GMMs = empirical_factory.determine_all_gmm(
+                    classdef.Fault(tect_type=classdef.TectType[tect_type]),
+                    str(im),
+                    model_dict,
+                )
+                for GMM_idx, (GMM, _) in enumerate(GMMs):
+                    # TODO: Check those models
+                    if GMM.name in ("K_20", "K_20_NZ", "ZA_06", "CB_14", "ASK_14"):
```
jasonmotha, line 126: I guess we can merge it in the interim, but really this "story" isn't complete till we have interpolation and all GMMs happy.

jasonmotha (changes requested):
Awesome work!

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
SnowNooDLe: Just a thought, shall we implement this in a different PR? :thinking:

jasonmotha: Should be fine in this PR as existing functionality shouldn't be affected.

joelridden (commented):
I'm happy, just waiting for new changes from recent comments

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
```diff
@@ -0,0 +1,248 @@
...
+        )
+
+        # Check to see if any site's Z1.0 or Z2.5 is NaN
+        if np.any(np.isnan(site_df.z1p0.values)):
+            raise ValueError("Z1.0 cannot be NaN")
+        if np.any(np.isnan(site_df.z2p5.values)):
+            raise ValueError("Z2.5 cannot be NaN")
+
+        # Filter the site/station that the distance_store actually has
+        sites = [
```
claudio525, line 108: I suggest getting the available stations from the `distance_store` (i.e. using `stored_stations`) and then using `np.isin` (or the pandas version). For larger number of stations using `np.isin` is significantly faster

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
claudio525: Maybe worth to add a print statement that those models are being skipped?

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
claudio525: https://github.com/ucgmsim/gmhazard/blob/d99b40b261d2c6f44c64b8859b3b9ee99d8ae42b/calculation/gmhazard_calc/gmhazard_calc/utils.py#L409

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
claudio525: @jasonmotha Assuming I'm not missing something here, I recon the way Tom has done it is good, it's nice and tidy imo. The IMDB is open for the for the minimal amount of time vs keeping it open (or opening it again). And it doesn't requires having to keep track of the IMDBs in a dictionary (i.e process and forget).

claudio525: @SnowNooDLe @jasonmotha From my understanding `common.write_metadata_and_close()` is a different approach. It assumes that all IMDBs are opened initially and then kept open throughout the script and then after processing `common.write_metadata_and_close()` is called to write metadata to all of them and then close them all (which was needed for the previous MPI version). 
Whereas what you have written above, only opens each IMDB when needed, writes the data (and metadata) and then closes it before moving onto the next, which imo is cleaner. 
If we are also going do tidy up, then I recon we should bring `calc_emp_flt.py` in line with what you have above, which I think is the only other usage of `common.write_metadata_and_close()`

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
```diff
@@ -0,0 +1,260 @@
...
+                                    rupture_df["rupture_name"].isin(
+                                        rupture_context_df["rupture_name"]
+                                    )
+                                ].index,
+                                inplace=True,
+                            )
+
+                            # Relabel the columns
+                            # PGA_mean -> PGA
+                            gmm_calculated_df.columns = [
```
jasonmotha, line 175: @claudio525 Do you think it's better to have one for loop than two list comprehensions?
I'm not sure which is cleaner.

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
```diff
@@ -0,0 +1,260 @@
...
+        args.site_source_db,
+        args.vs30_file,
+        args.output_dir,
+        z_ffp=args.z_file,
+        ims=args.im,
+        psa_periods=args.periods,
+        model_dict_ffp=args.model_dict,
+        suffix=args.suffix,
+    )
+    print(f"Finished in {(time.time() - start) / 60}")
```
jasonmotha, line 260: Put a unit on this and some float rounding would be good. ":.2f"

jasonmotha (changes requested):

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
joelridden: I personally think I'd prefer one for loop?

SnowNooDLe: I was in that dilemma. However, even if we support all IMs we have, columns will be smaller than 200 and apparently, list comprehension seems faster in this kind of simple case as well. https://stackoverflow.com/questions/44666215/single-list-iteration-vs-multiple-list-comprehensions

Hence, I chose list comprehensions over one for loop.

jasonmotha: Magic. :exploding_head: 

Since that's the case, I'd suggest making a function to remove the duplicate code.

Should just need the df, a search string, and an optional suffix

claudio525: Does it always have the the `_mean` and `std suffix`? If thats the case you can also do it with np.char functionality (i.e. vectorise it)
Also, instead of splitting the whole string, you can just use `rsplit` and set maxsplit then you don't need a join.

In terms of for loop vs list comprehension, I think in this case either or seems fine, I would probably use list comprehensions.

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
```diff
@@ -0,0 +1,260 @@
...
+                                for col in gmm_calculated_df
+                            ]
+
+                            # Write an im_df to the given station/site
+                            imdb.add_im_data(
+                                site,
+                                gmm_calculated_df.loc[
+                                    :,
+                                    # Only mean and std_Total are needed
+                                    ~gmm_calculated_df.columns.str.endswith(
```
claudio525, line 195: Instead of masking here, you could drop those two columns after you get gmm_calculated_df back from the wrapper. Just a thought.

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
claudio525: Actually, instead of using split, you could also use rstrip, e.g. for the `mean` case:
```
gmm_calculated_df.columns = np.char.rstrip(gmm_calculated_df.columns.values.astype(str), "_mean")
```
is all you need I think? (I didn't test it, so no guarantees haha)

SnowNooDLe: @claudio525 CC(@jasonmotha) Yeah, it does always have `_mean` and `_std_suffix` and managed to update those with magic numpy methods, so we probably don't need to create a small helper function as one is using `rstrip` and another one is using `replace`

tools/gmhazard_scripts/db_creation/empirical_db/new_calc_emp_ds.py
SnowNooDLe: Tried to look up the ways of dropping columns with multiple conditions but it all ended up masking was the simpler/only option I could find.

But minor update, it turns out pandas' string methods `endswith` only works with one pattern, so replace it with `contains`

jasonmotha (approved):

joelridden (approved):

claudio525 (approved):

PR #62, opened 2022-04-01
Ensemble load fix

claudio525:
- Added IM lazy loading to both Ensemble and IMEnsemble

### DIFF
--- a/calculation/gmhazard_calc/gmhazard_calc/gm_data/IMEnsemble.py
+++ b/calculation/gmhazard_calc/gmhazard_calc/gm_data/IMEnsemble.py
@@ -18,7 +18,7 @@ class IMEnsemble:
 
     Parameters
     ----------
-    ims: sequence of IMType or IMType
+    im_types: sequence of IMType or IMType
         The IMs this IM-Ensemble supports
     ensemble: Ensemble
         The parent ensemble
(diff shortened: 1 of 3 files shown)


calculation/gmhazard_calc/gmhazard_calc/gm_data/IMEnsemble.py
```diff
@@ -147,3 +141,29 @@ def __load_rupture_df(self):
...
+        )
+
+    def __load_stations(self):
+        branch_stations = set.intersection(
+            *[set(branch.stations) for branch in self.branches_dict.values()]
+        ).intersection(self.ensemble.stations_ll_df.index)
+
+        self._stations = self.ensemble.stations_ll_df.loc[
+            branch_stations
+        ].sort_index()
```
joelridden, line 169: No newline EOF, maybe this file needs to be blacked?

joelridden (commented):
Looks good to me

jasonmotha (dismissed):
Pending formatting changes

calculation/gmhazard_calc/gmhazard_calc/gm_data/IMEnsemble.py
```diff
@@ -147,3 +139,27 @@ def __load_rupture_df(self):
                 self._rupture_df = self._rupture_df.loc[
                     ~self._rupture_df.index.duplicated()
                 ]
+
+    def __load_ims(self):
+        if IMType.pSA in self._im_types:
+            self._ims = set(list(self.branches[0].ims))
```
SnowNooDLe, line 145: Just a thought to be consistent,
Shall we have the following code instead? so we can safely say `self._ims` is always a list
```python
list(set(list(self.branches[0].ims)))
```

SnowNooDLe (commented):
Just one comment wants to be checked.

and I triggered another benchmark test for this PR as well.

calculation/gmhazard_calc/gmhazard_calc/gm_data/IMEnsemble.py
SnowNooDLe: But at the same time, the set is still iterable and we will polish this `self._ims` again to be `np.array` so probably not worth casting it.

SnowNooDLe (approved):

joelridden (approved):

PR #63, opened 2022-04-05
Update the logic on `get_sites` endpoint 

SnowNooDLe:
# Update the logic on `get_sites` endpoint 

Instead of a hacky way of creating an Ensemble to get Latitude and Longitude information,
project_id.yaml now includes the Lat&Lng information. So we read those from yaml file instead.

## Current version
image

## Updated version
image

## No differences when comparing within PyCharm 
image


jasonmotha (dismissed):

apis/project_api/project_api/api/project.py
```diff
@@ -53,29 +53,22 @@ def get_available_sites():
...
     project = utils.get_project(version_str, project_id)
-
-    ## Hack: Have lat/lon in the project config file later
-    ensemble = sc.gm_data.Ensemble(project_id, config_ffp=project.ensemble_ffp)
+    # Read the project_id yaml which includes lat&lon information
+    with open(
+        Path(project.ensemble_ffp.rsplit("/", maxsplit=1)[0]) / f"{project_id}.yaml"
+    ) as f:
+        config = yaml.safe_load(f)
+    locations = config["project_parameters"]["locations"]
```
joelridden, line 61: I think this can be simplified by just getting project.config?
Then you can get the locations with
`locations = project.config["project_parameters"]["locations"]`

SnowNooDLe: Even better, thanks!

joelridden (approved):

jasonmotha (approved):

claudio525 (approved):

PR #71, opened 2022-05-23
Fix the NZS1170.5 test data and test itself

SnowNooDLe:
# Fix the NZS1170.5 test data and test itself

- Update the logic to check both radius 0 and 2.
- Update the test data as the soil class is now `soft_or_deep_soil` when Vs30 < 300 instead of < 350.


jasonmotha:
Do you need to include the vs30 code change too?

SnowNooDLe:
> Do you need to include the vs30 code change too?

This PR goes into the feedback repo which does have that change haha

joelridden (approved):

jasonmotha (approved):

PR #72, opened 2022-05-25
Projects - Replace labels with the CONSTANT variables

SnowNooDLe:
# Projects - Replace labels with the CONSTANT variables

This PR focuses on the Projects tab's labels.

Upcoming PR will replace labels in the Non-Projects tab.

### DIFF
--- a/frontend/src/constants/Constants.js
+++ b/frontend/src/constants/Constants.js
@@ -125,7 +125,6 @@ export const ENSEMBLE_BRANCHES = "Ensemble branches";
 export const FAULT_DISTRIBUTED_SEISMICITY_CONTRIBUTION =
   "Fault/distributed seismicity contribution";
 export const DOWNLOAD_DATA = "Download data";
-export const RETURN_PERIOD_WITH_UNIT = "Return Period (years)";
 export const GET_BUTTON = "Get";
 export const GMS_IM_DISTRIBUTIONS_PLOT = "IM Distributions";
 export const GMS_CONDITIONING_IM_NAME = "Conditioning IM Name";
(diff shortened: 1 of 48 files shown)


frontend/src/constants/Constants.js
```diff
@@ -110,19 +112,114 @@ export const INTERMEDIATE_API_ALL_PRIVATE_PROJECTS_ENDPOINT =
...
+export const PLACEHOLDER_NOT_APPLICABLE = "N/A";
+export const PLACEHOLDER_SELECT_COMPONENT = "Select component...";
+
+// Units in labels
+export const SECONDS = "(s)";
+export const KILOMETRE = "(km)";
+export const METRE_PER_SECOND = "(m/s)";
+export const YEARS = "(years)";
+export const PERCENTAGE = "(%)";
+export const G_FORCE = "(g)";
```
jasonmotha, line 222: I would have these with a suffix of _UNIT (I presume these compile out for the webapp side itself?)

jasonmotha (approved):

frontend/src/constants/Constants.js
SnowNooDLe: Yeah, these will be with a suffix of _UNIT as you pointed out. (In upcoming PR)
I just realized that in the non-projects tab with Mw, a symbol for Magnitude, because `MAGNITUDE` is already reserved. Hence, I had to add a suffix, and might as well add this suffix to other units to keep the consistency :+1:

joelridden (approved):

claudio525 (approved):

PR #73, opened 2022-05-26
Non-projects' label updates

SnowNooDLe:
# Non-projects' label updates

**Wait for the previous PR to go through first**

It is focused on the non-projects tab.
- Any labels that require more than a combination of two constants or long labels are replaced with just one constant label.
E.g.,
image
- `PermissionConfig` and `ProjectCreate`
You may see some changes within those files, but they are because of the labels used in multiple places. It is because those labels are used in various places. Once we get to the point where we decide to keep those, we can certainly update them too.
- Previous PR's feedback and pre-reserved constant, `MAGNITUDE`, units' constants now have a suffix, `_UNIT`
- Minor bug fix
Unlike the projects tab, for the non-projects, we have to make sure these are not `null`
image


frontend/src/constants/Constants.js
```diff
@@ -214,12 +277,17 @@ export const PLACEHOLDER_NOT_APPLICABLE = "N/A";
...
-export const METRE_PER_SECOND = "(m/s)";
-export const YEARS = "(years)";
-export const PERCENTAGE = "(%)";
-export const G_FORCE = "(g)";
+export const SECONDS_UNIT = "(s)";
+export const KILOMETRE_UNIT = "(km)";
+export const METRE_PER_SECOND_UNIT = "(m/s)";
+export const YEARS_UNIT = "(years)";
+export const PERCENTAGE_UNIT = "(%)";
+export const G_FORCE_UNIT = "(g)";
```
jasonmotha, line 285: I'd just call this gravity unit. Force and acceleration are different but this could be confusing.

frontend/src/constants/Constants.js
```diff
@@ -214,12 +277,17 @@ export const PLACEHOLDER_NOT_APPLICABLE = "N/A";
...
+export const MAGNITUDE_UNIT = "(Mw)";
```
jasonmotha, line 286: Magnitude is dimensionless so doesn't have a unit but does have a symbol: Mw

joelridden (approved):
Seems fine to me. Jason's comments though make sense to do

claudio525 (dismissed):

frontend/src/components/Hazard/GMS/GMSForm.js
```diff
@@ -803,7 +803,8 @@ const GMSForm = () => {
                       </tr>
                       <tr>
                         <th scope="row">
-                          {CONSTANTS.RUPTURE_DISTANCE} {CONSTANTS.KILOMETRE_UNIT}
+                          {CONSTANTS.RUPTURE_DISTANCE}{" "}
+                          {CONSTANTS.KILOMETRE_UNIT}
```
joelridden, line 807: I think {" "} can just be a space like on line 785?

jasonmotha (approved):

joelridden (approved):

PR #74, opened 2022-05-29
Scenario Calculation change

joelridden:
Now for mu, using the weighted average.
For 16th and 84th, using the following formula for getting variance.
Standard deviation from that and 1 above and below is shown for 16th and 84th.
image
(Just no squared on the weight in the 2nd part of this equation)

Before: (Different faults, but you get the idea)
image
After: (Different faults, but you get the idea)
image


calculation/gmhazard_calc/gmhazard_calc/scenario/scenario.py
```diff
@@ -36,78 +36,37 @@ def run_ensemble_scenario(
...
-    weights = np.take_along_axis(weights, sort_ind, 1)
-
-    # Inverse CDF lookup
-    cdf_x, cdf_y = im_branch_ruptures, np.cumsum(weights, axis=1)
-    x_values = sha_calc.shared.query_non_parametric_multi_cdf_invs(
-        [0.16, 0.5, 0.84], cdf_x, cdf_y
+    # Calculate mu
+    mu = sum(
+        cur_scenario.branch.weight * cur_scenario.mu_data
+        for cur_scenario in scenario_branches
```
jasonmotha, line 44: I'm pretty sure this should be log mean. Probably worth checking.

frontend/src/components/common/Scenarios/ScenarioPlot.js
```diff
@@ -44,7 +46,7 @@ const ScenarioPlot = ({ scenarioData, scenarioSelectedRuptures, extra }) => {
           type: "scatter",
           mode: "lines",
           line: { color: deafultColours[colourCounter % 10] },
-          name: `Scenario ${curRup} [16, 50, 84 %ile]`,
```
SnowNooDLe, line 47: It was me who didn't remove `Scenario` but like every other plot, I think we can safely remove the `Scenario`
Like Claudio said the other day, users are already on the Scenario page and no other plots(Hazard Curve, UHs and GMS), we do not say `Hazard Curve - something`.

calculation/gmhazard_calc/gmhazard_calc/scenario/scenario.py
SnowNooDLe: Yeah, I was confused too...

Does this `mu` indicate the weighted average?

frontend/src/components/common/Scenarios/ScenarioPlot.js
```diff
@@ -44,7 +46,7 @@ const ScenarioPlot = ({ scenarioData, scenarioSelectedRuptures, extra }) => {
...
+          name: `Scenario ${curRup} [mean and 16<sup>th</sup>, 84<sup>th</sup> percentile]`,
```
claudio525, line 49: Imo these should also be renamed, since its no longer the 16th and 84th. Maybe something like $\mu \pm \sigma$?

Not really sure on this though haha

jasonmotha:
https://github.com/ucgmsim/gmhazard/pull/74#discussion_r885124226

This makes sense. It means that the IM values would be consistent with what we do in the hazard curves. 
Joel and I looked at whether the hazard branch combining used ln mean or not and it appeared to not but if the IM_values were already in log space that would make sense.

calculation/gmhazard_calc/gmhazard_calc/scenario/scenario.py
jasonmotha: This makes sense. It means that the IM values would be consistent with what we do in the hazard curves.
Joel and I looked at whether the hazard branch combining used ln mean or not and it appeared to not but if the IM_values were already in log space that would make sense.

frontend/src/components/common/Scenarios/ScenarioPlot.js
jasonmotha: Isn't /pm 1 /sigma the equivalent to the 16th/84th percentile?

SnowNooDLe: Are they no longer 16th/84th percentiles?

frontend/src/components/common/Scenarios/ScenarioPlot.js
```diff
@@ -11,8 +11,8 @@ const ScenarioPlot = ({ scenarioData, scenarioSelectedRuptures, extra }) => {
   if (scenarioData !== null && !scenarioData.hasOwnProperty("error")) {
     const percentileData16 =
       scenarioData["ensemble_scenario"]["percentiles"]["16th"];
-    const percentileData50 =
-      scenarioData["ensemble_scenario"]["percentiles"]["50th"];
+    const mu_data =
```
SnowNooDLe, line 14: `muData` haha

frontend/src/components/common/Scenarios/ScenarioPlot.js
```diff
@@ -39,20 +39,20 @@ const ScenarioPlot = ({ scenarioData, scenarioSelectedRuptures, extra }) => {
     });
 
     let colourCounter = 0;
-    for (let [curRup, curData] of Object.entries(percentileData50)) {
+    for (let [curRup, curData] of Object.entries(mu_data)) {
       if (scenarioSelectedRuptures.includes(curRup)) {
         scatterObjs.push({
           x: x_values,
```
SnowNooDLe, line 45: This just caught my eyes as well.
Can you also rename it to `xValues`?

SnowNooDLe (approved):
Looks good to me, just two renaming suggestions.

SnowNooDLe (dismissed):

jasonmotha (approved):

jasonmotha (approved):

SnowNooDLe (approved):

claudio525 (approved):

PR #79, opened 2022-06-20
Project gen UHS fix

SnowNooDLe:
# Project gen UHS fix

Instead of filling the whole `np.nan`, fill when needed.

So that way, on the frontend, check each Annual exceedance data and display the data that does not include `np.nan`


joelridden (approved):

jasonmotha (dismissed):
It's okay, but I don't really like temporary array variables called `temp`

happy to merge as is as I don't have any constructive suggestions.

calculation/gmhazard_calc/gmhazard_calc/uhs/uhs.py
```diff
@@ -176,15 +176,14 @@ def __get_pSA_values(
...
-        try:
-            pSA_branch_values.append(
-                [
-                    cur_branch.exceedance_to_im(cur_exceedance)
-                    for cur_branch in hazard_branches.values()
-                ]
-            )
-        except exceptions.ExceedanceOutOfRangeError as ex:
-            pSA_branch_values.append(np.full(len(hazard_branches), np.nan))
+        temp = []
```
claudio525, line 179: Agree with Jason, `temp` is not a good name, maybe `cur_pSA_branch_values` instead?

jasonmotha (approved):

joelridden (approved):

claudio525 (approved):

PR #87, opened 2022-07-06
Scenario download with rupture metadata

SnowNooDLe:
# Scenario download with rupture metadata

Scenario download now includes the rupture metadata.

image

### DIFF
--- a/apis/project_api/project_api/api/scenario.py
+++ b/apis/project_api/project_api/api/scenario.py
@@ -97,7 +97,7 @@ def download_ens_scenario():
     ensemble_scenario = sc.scenario.EnsembleScenarioResult.load(
         project_dir / "results" / station / im_component / "scenario",
     )
-    # Load metadata (Source contribution)
+    # Load Source contribution
     source_contribution = utils.load_scenario_source_contribution(
         project_dir,
         project_id,
@@ -107,7 +107,7 @@ def download_ens_scenario():
             sc.scenario.filter_ruptures(ensemble_scenario).to_dict()["mu_data"].keys()
         ),
     )
-    breakpoint()
+
     with tempfile.TemporaryDirectory() as tmp_dir:
         zip_ffp = au.api.create_scenario_download_zip(
             ensemble_scenario, source_contribution, tmp_dir, prefix=f"{project_id}",
--- a/apis/project_api/project_api/api/scenario.py
+++ b/apis/project_api/project_api/api/scenario.py
@@ -110,7 +110,7 @@ def download_ens_scenario():
 
     with tempfile.TemporaryDirectory() as tmp_dir:
         zip_ffp = au.api.create_scenario_download_zip(
-            ensemble_scenario, rupture_metadata, tmp_dir, prefix=f"{project_id}",
+            ensemble_scenario, rupture_metadata, tmp_dir
         )
 
         return flask.send_file(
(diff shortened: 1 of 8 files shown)


apis/project_api/project_api/api/scenario.py
```diff
@@ -92,20 +92,25 @@ def download_ens_scenario():
...
+    rupture_metadata = utils.load_scenario_rupture_metadata(
+        project_dir,
+        project_id,
+        station,
+        sc.im.IMComponent[im_component],
+        list(
+            sc.scenario.filter_ruptures(ensemble_scenario).to_dict()["mu_data"].keys()
+        ),
+    )
+    # breakpoint()
```
joelridden, line 110: Remove commented breakpoint

apis/project_api/project_api/api/scenario.py
```diff
@@ -92,20 +92,25 @@ def download_ens_scenario():
...
     with tempfile.TemporaryDirectory() as tmp_dir:
         zip_ffp = au.api.create_scenario_download_zip(
-            ensemble_scenario, tmp_dir, prefix=f"{project_id}",
+            ensemble_scenario, rupture_metadata, tmp_dir
```
joelridden, line 113: Does this still need the prefix as a parameter?

SnowNooDLe: Yup, we do not use the prefix for now but looks like it will be used at some points as every download function actually does have a prefix param but it's optional.

apis/project_api/project_api/utils.py
```diff
@@ -194,9 +194,20 @@ def load_scenario_metadata(
     data_dir = list(station_data_dir.glob(f"disagg_{ims[0].file_format()}*"))[0]
     metadata_df = pd.read_csv(list(data_dir.glob("*_metadata.csv"))[0], index_col=0)
 
+    ensemble_disagg_result = gc.disagg.EnsembleDisaggResult.load(data_dir)
+    merged_metadata_df = ensemble_disagg_result.total_contributions_df.merge(
+        metadata_df, how="left", left_index=True, right_index=True
+    )
+    # Drop unnecessary columns and rows
+    merged_metadata_df.drop(labels=["contribution", "epsilon"], axis=1, inplace=True)
+    merged_metadata_df.drop(labels=["distributed_seismicity"], axis=0, inplace=True)
```
jasonmotha, line 203: generally you are better doing `merged_metadata_df = ... ` without the inplace as it is faster. The inplace is really only for when you can't make a copy in the memory available.

In also saying that, you'd be better off doing the drop's as "one line" - I think.

apis/project_api/project_api/utils.py
```diff
@@ -194,9 +194,20 @@ def load_scenario_metadata(
...
+
+    # Swap columns
+    new_order = ["rupture_name", "annual_rec_prob", "magnitude", "rrup"]
```
claudio525, line 206: Not sure if a variable is needed here? But thats preference issue. If you want to keep the variable then I do think naming it `column_order` is more appropriate

apis/project_api/project_api/utils.py
SnowNooDLe: drop as `one line` :exploding_head: 
The beauty of Python...

apis/project_api/project_api/utils.py
SnowNooDLe: Yea, fair. Even with me, I don't think it's necessary to declare a variable.

joelridden (approved):

claudio525 (approved):

jasonmotha (approved):

PR #90, opened 2022-07-20
Fixed rupture selection bug

claudio525:
- A realisation should be based on a single rupture, NOT one rupture per IMi of the realisation.
- Also updated dataframes to have string column names instead of IM objects


SnowNooDLe (approved):

joelridden (approved):

jasonmotha (approved):
Looks fine to me. I don't particularly understand the behaviour that was incorrect. I presume it was something to do with Simulation based GM Selection?

PR #93, opened 2022-08-02
Update react-script

SnowNooDLe:
# Update react-script

When we noticed critical vulnerability, I remember not to do `npm audit fix --force` just because seeing the vulnerability as it could break the entire application due to a change in the major version. (In our case, react-script bumped up from 3.4.3 to 5.0.1)

I did some research and concluded that it's better to do `npm audit fix` regularly if needed but just be careful when it comes to `npm audit fix --force` as it's likely to update to the latest stable version. (But the latest stable does not mean it's guaranteed to work with our existing product.)

However, the change was only from `react-script`, which is there to do something like `create-react-app`. I also tested to deploy the test version of GMHazard under /gmhazard_dev, and it works well.

Hence, I am quite confident that we can put this into the main branch.


jasonmotha (approved):
You've tested it right?

SnowNooDLe:
> You've tested it right?

Yup, I actually deployed like a dev version to test haha
https://quakecoresoft.canterbury.ac.nz/gmhazard_dev/project

joelridden (approved):
Agreed

claudio525 (approved):

PR #105, opened 2023-03-16
Get all rrup data from site source db's

joelridden:
Used in cybershake Investigation for bulk rrup data while calculating empirical data for an archived version of cybershake


joelridden:
Not needed anymore as this functionality is now moved directly inside cybershake investigation

PR #109, opened 2023-09-01
Spatial hazard mods

claudio525:


calculation/spatial_hazard/spatial_hazard/im_dist.py
```diff
@@ -1,6 +1,6 @@
 import pickle
 from pathlib import Path
-from typing import Sequence, Callable, List, Dict, Tuple
+from typing import Sequence, Callable, List, Dict, Tuple, Union
```
joelridden, line 3: Don't think Union is being used?

joelridden (approved):

